PR #5, opened 2017-02-22
LATimes.com anti-ad block fix

lukemulks:
Fixes the overlay that blocks content unless the user whitelists


bridiver:
++ fast work @lukemulks

PR #10, opened 2017-03-03
correction latimes.com native ads rules

lukemulks:
rule that was submitted and merged was missing a character. all others were correct and are functioning as expected. 

rule that has the issue: 
||adserve.postrelease.com^$script,image,domain=latimes.com

correct version: 
||jadserve.postrelease.com^$script,image,domain=latimes.com

tested in about:adblock, once corrected, it fully resolves.


bbondy:
++

Issue #8, opened 2017-03-02
LAtimes.com - filter rules to fix new native ads


lukemulks:
3 different flavors of new, native ads discovered as of this morning on latimes.com - being tracked and investigated here: https://github.com/brave/browser-laptop/issues/7454

These are different and new, unrelated to recent PR that was merged last week. 

2 different versions of "sponsored content" ads discovered this morning on Windows, which the rules below have confirmed resolving when tested in `about:adblock`. 

```
||aggrego.org^$script,image, domain=latimes.com
||adserve.postrelease.com^$script,image, domain=latimes.com
||troncdata.com^$script,image, domain=latimes.com
||polarmobile.com^$script,image, domain=latimes.com
||ntv.io^$script,image, domain=latimes.com
||gigya.com^$script,image, domain=latimes.com
```

A 3rd issue has been reported from MacOS, but has not yet been reproduced in Windows. Still investigating that. 

Will submit a PR for adding the rules above to solve for 2/3, and will follow up here if the third issue is also resolved once I'm able to verify from MacOS.

lukemulks:
PR submitted: https://github.com/brave/adblock-lists/pull/9

lukemulks:
Confirming that after testing on iOS, no longer seeing the ads display that the rules above resolve. 

The third item I was unable to repro was confirmed as a shield-down observation, so considering the PR above the fix for the issues reported today.

lukemulks:
@bbondy confirming that all the rules merged above are functioning as expected in 0.13.5, except for one which I made an error on. 

I've corrected the error, and submitted this PR to resolve. https://github.com/brave/adblock-lists/pull/10

Apologies for the oversight. Tested the update with the corrected version in about:adblock, and it fully resolves the issue.

matt2000:
Hi, I'm the developer for troncdata.com (an internal service of LATimes / Tronc) and I want to let you know that it does not serve advertising at all, native or otherwise. Can you please remove it from you bock list? Thanks.

lukemulks:
We're going to have to agree to disagree in this instance.  

I am going to **_strongly object_** to removing troncdata from the blocklist. 

I've reversed the implementation, and `troncdata` appears to be used from the 1st party level with an ad widget, invoking an `infusion` script function with nested timing sequences for matching page data and user IDs to delayed ad calls that bind and replace to an html `<aside>` container in the right rail. `data-state` and other `data-*` attributes are used in a clever way in this implementation.

[...]

PR #13, opened 2017-03-09
#8: troncdata.com is content, not advertising.

matt2000:


bbondy:
cc @lukemulks for thoughts

lukemulks:
@matt2000 @bbondy the domain was being used for native ads that were designed to serve with ad blockers enabled. Not too uncommon for this type of a case. 

That said, I will review and confirm if this domain has multi-purpose use (content and ads), and see if a more granular rule can be created to only filter for troncdata.com for ad and tracking-related requests. 

Will update here following review.

lukemulks:
@matt2000 @bbondy I'd advise we keep the filter rule in place that blocks troncdata.com.  

@matt2000 can you confirm if issues were reported to you that specifically called out webcompat issues with this blocked? 

IMO we should keep this blocked, unless there are specific reports of webcompat issues, along with supporting info, that show that blocking troncdata.com is causing issues. 

From what I've reviewed today, it appears that the troncdata.com is a 1p stub proxy for passing data used for native ad placements, and that user data is passed that we should be blocking. 

Definitely open to reassessing, if more info can be provided to support removing the filter. LATimes.com is having several issues reported with other ad blockers at the moment. For example, the paywall prompt is back for uBO and ABP users, which we're blocking thanks to the latest updates. 

Additionally, there are some strange exceptions made for Rubicon and other advertisers on the Easylist that I don't recall seeing in the past. They stick out like a sore thumb in network traffic w/ABP right now. I'm not talking about the Easylist Exceptions for Acceptable Ads either, I'm talking about the Easylist.to list. 

**_For example,_** I do not know what could possibly justify this exception that is made on the Easylist, and there appears to be no record of it in the ABP or Easylist forum: 
`@@||rubiconproject.com/header/7476.js$script,domain=baltimoresun.com|boston.com|calgaryherald.com|capitalgazette.com|carrollcountytimes.com|chicagotribune.com|courant.com|dailypress.com|deathandtaxesmag.com|edmontonjournal.com|edmunds.com|financialpost.com|gofugyourself.com|hearthhead.com|infinitiev.com|latimes.com|leaderpost.com|lolking.net|mcall.com|montrealgazette.com|nasdaq.com|nationalpost.com|orlandosentinel.com|ottawacitizen.com|ranker.com|sandiegouniontribune.com|saveur.com|sherdog.com|spin.com|sporcle.com|stereogum.com|sun-sentinel.com|theprovince.com|thestarphoenix.com|timeanddate.com|tmn.today|vagazette.com|vancouversun.com|vibe.com|windsorstar.com`

**^^imo, an exception for Rubicon is nuts.** Rubicon is pretty horrible. Rubicon Fastlane is deliberately designed to run ahead of other ad calls, and this is pretty suspicious to see IMO. We block this, uBO as well, but ABP is not blocking this. 

Another strange discovery in the Easylist...
`@@||s.ntv.io/serve/load.js$domain=baltimoresun.com|boston.com|calgaryherald.com|capitalgazette.com|carrollcountytimes.com|chicagotribune.com|courant.com|dailypress.com|deathandtaxesmag.com|edmontonjournal.com|edmunds.com|financialpost.com|gofugyourself.com|hearthhead.com|infinitiev.com|latimes.com|leaderpost.com|lolking.net|mcall.com|montrealgazette.com|nasdaq.com|nationalpost.com|orlandosentinel.com|ottawacitizen.com|ranker.com|sandiegouniontribune.com|saveur.com|sherdog.com|spin.com|sporcle.com|stereogum.com|sun-sentinel.com|theprovince.com|thestarphoenix.com|timeanddate.com|tmn.today|vagazette.com|vancouversun.com|vibe.com|windsorstar.com`

We just applied rules to block `ntv.io `^^ after discovering that the domain was being used for native ads and tracking last week. After seeing the above exception made to the Easylist, I am even more wary of removing troncdata.com, as it was discovered when we were investigating native ads and tracking on latimes.com. 

Again, open to reassessing if more info is provided, but I strongly suggest that unless there is strong justification for removing the filter rule, we keep it in place.

lukemulks:
@bbondy follow-up after review noted here https://github.com/brave/adblock-lists/issues/8

matt2000:
Copying here my follow-up from #8 since it hasnt received a response there:

@lukemulks Your analysis is incorrect, and I'd like to help clarify the purpose of this widget. I wrote widget.js and the backend service at recommend.troncdata.com that it communicates with. I'm happy to be 100% transparent about what's happening here. (As you can see, the script is not even minimized, let alone obfuscated.) I work directly for Tronc / LA Times, and have since July 2016. (I.e., I'm not a contractor or third-party service provider. I collaborate with the newsrooms on this project, not ad sales.)

There is no connection to gigya at all, and it's not clear to me how you got the idea that there was. The <script> blocks in your 3rd screen shot are all completely independent. I don't even know what "mediaconductor" is (many parts of the page are owned by different departments), but you can clearly see it's not used by my script.

I directly and personally control troncdata.com, not gigya or any ad provider. I can think of various ways to prove that to you, if needed. (E.g., adding DNS records or custom HTTP responses, although I don't have convenient access to modify the domain WHOIS, but could arrange it if that would satisfy you.)

The ID used is generated by Ensighten, and is merely a convenient identifier for matching users across our analytics platforms.

We use the User UUID only to track what user's have read so as to not show them content they've already seen. In the near future, we expect to use it to recommend to them content that is similar to content they've read, or of interest to other users who've read similar content. (Collaborative filtering, etc.)

This widget does not share data with any third-party. In fact, it doesn't even persist it's own data beyond the lifetime of an individual server thread. (Long-term user behavior is gathered from regular web analytics, so we only care about real-time activity, and update form the analytics data when it becomes available in batch.)

We're currently A/B testing against the original <aside> block. The fact you see the response {"recsys": "control"} just means you're in the control group (based on UUID, along wiht 70% of users as of today) and so the HTML is unmodified. As you read the JS, you know how to opt yourself into the "B" group if you want to see what the returned mark-up looks like for users in the test-group. (It's identical, except for the articles that are selected to appear.)

We'll probably ultimately be doing this on the first-party domain, so there's really no point in cluttering your list. I don't expect you to believe me, but you should have to, because the purpose of this widget is not to serve advertising. It's only for content recommendation, i.e., legitimate news articles by our journalists. Please spend your time knocking down the Taboola's and Outbrain's of the world. We're just trying to help users find quality journalism content that interests them, from a real news agency.

Please let me know if there are any other questions I can answer to set your concerns at ease.

matt2000:
@lukemulks @bbondy How can I help here? My goal is for Brave to be able to tell the difference between Intrusive Advertising (bad!) vs Content (good!), and between Tracking for Ad-Targeting and Surveillance (bad!!) vs Tracking for Experience Optimization (good!!). How can we solve this?

matt2000:
To confirm: Troncdata.com does not serve ads. Period. Whatever methodology you used to decide that it does, is "buggy." I want to help you fix that methodology.

PR #26, opened 2017-09-19
Block the mining of jsecoin

jonathansampson:
Here's a sample of what is added to the publisher's site:

```js
<script type="text/javascript">
(function() {
var trackingurl = 'https://load.jsecoin.com/server/load/3864/sampsonblog.com/0/0/';
var d=document, j=d.createElement('script'), s=d.getElementsByTagName('script')[0];
j.type='text/javascript'; j.async=true; j.defer=true; j.src=trackingurl; s.parentNode.insertBefore(j,s);
})();
</script>
```

No indication is given that the location will ever be anything other than _load.jsecoin.com_.

### DIFF
--- a/coin-miners.txt
+++ b/coin-miners.txt
@@ -1,1 +1,2 @@
 ||coin-hive.com$third-party
+||jsecoin.com$third-party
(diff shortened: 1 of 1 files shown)


bbondy:
I'll accept either way but would you happen to have a test site to verify it's working?

bbondy:
Thanks @jonathansampson !

PR #40, opened 2019-01-11
added exception rule to allow reddit impression pixel, fixes #39

### LINKED ISSUE #39: Exceptions for two additional reddit requests
Reddit has connected to mention that the following requests are currently being blocked, that they use for 1st party ads that they run. 

These are the requests that Reddit has requested we make exceptions for: 

 - Request: https://www.redditstatic.com/desktop2x/js/ads.js 
 - Explanation: redditstatic.com is a domain we use to host many static assets
 - Request: https://alb.reddit.com/i.gif 
 - Explanation: This is our impression pixel, so even though reddit native ads are showing on brave right now we're not getting paid for any of them because this pixel can't fire.

Two possible exception filters: 

For 1: @@||redditstatic.com^
For 2: @@||alb.reddit.com^

These would be appended to this section of the unbreak list (should have been added initially, were not discovered until after the fact): 

https://github.com/brave/adblock-lists/blob/master/brave-unbreak.txt#L88

snyderp:
adds an exception rule to allow reddit's impression pixel through


snyderp:
@lukemulks looks like this got overlooked (likely my fault).  Do you know if its still needed?  If you're not sure, or the reddit folks aren't still asking for it, i'll just close the PR to keep things tidy :)

PR #45, opened 2019-02-24
Move hardcoded exceptions rules to more specific rules and unify exception handling

bbondy:
Related issue: [brave/brave-browser#3475](https://github.com/brave/brave-browser/issues/3475)'
Related PR: https://github.com/brave/brave-core/pull/1770

This is the first part of a series of work that will be done within the next few weeks. 
It unifies exception handling in one place.

In a later PR, we will be adding metadata to filter rules, and exposing options to control those rules.
It will be opt-out options at first, and then a later iteration will try going to opt-in with UX to notify the user that things are broken.

See this issue for a description of the follow up work that will be done:
https://github.com/brave/brave-browser/issues/3475

This is specifically for:
- Facebook login buttons
- Facebook embed posts
- Twitter embeds

Note that this does not change existing functionality, it just moves a hardcoded c++ list to a more unified place where we can iterate on to add options to exclude them.

For a test plan, see the linked issue above.

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -88,6 +88,7 @@
 ! Allow 1P piwik since it's no different than server logs
 @@https://analytics.*/piwik.$~third-party
 ! Fully block Facebook everywhere just like our Disconnect blocking but unbreak logins like from quora.com and twitch.tv
+! Note that options will be added to exclude these filters soon.
 ||facebook.com$third-party
 ||facebook.net$third-party
 ||staticxx.facebook.com$third-party
@@ -97,3 +98,11 @@
 @@||www.facebook.com/connect
 @@||staticxx.facebook.com/connect/
 @@||graph.facebook.com/
+! Block all twitter.com third-party but allow embedded tweets
+! Note that options will be added to exclude these filters soon.
+||twitter.com$third-party
+@@||platform.twitter.com/
+@@||syndication.twitter.com;
+||twimg.com$third-party
+@@||pbs.twimg.com/
+@@||cdn.syndication.twimg.com/
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -98,11 +98,19 @@
 @@||www.facebook.com/connect
 @@||staticxx.facebook.com/connect/
 @@||graph.facebook.com/
+! Block fbcdn.net everywhere but allow Facebook embeds
+||fbcdn.net$third-party,domain=~facebook.com
+@@||staticxx.facebook.com/
+@@||xx.fbcdn.net/
+@@||www.facebook.com/*/plugin
+@@||www.facebook.com/plugins/
+@@||www.facebook.com/rsrc.php
+@@||www.facebook.com/ajax/bz
 ! Block all twitter.com third-party but allow embedded tweets
 ! Note that options will be added to exclude these filters soon.
 ||twitter.com$third-party
 @@||platform.twitter.com/
 @@||syndication.twitter.com;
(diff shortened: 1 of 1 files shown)


bbondy:
Pls don't merge if approved.

Note that this blocks more strictly than disconnect but I didn't see an issue caused by it.  
We'd have to add an extra allow exception I think to make it parity.

brave-unbreak.txt
```diff
@@ -88,3 +87,27 @@
...
+@@||connect.facebook.net/*/sdk.js$script
+@@||facebook.com/connect/
+@@||www.facebook.com/connect
+@@||staticxx.facebook.com/connect/
+@@||graph.facebook.com/
+! Block fbcdn.net everywhere but allow Facebook embeds
+||fbcdn.net$third-party,domain=~facebook.com
+@@||staticxx.facebook.com/
+@@||xx.fbcdn.net/
+@@||www.facebook.com/plugins/post.php?href=
```
fmarier: There are [a few more URLs](https://github.com/brave/brave-browser/issues/3241#issuecomment-465825247) for Facebook embeds:

- `www.facebook.com/plugins/video.php`
- `www.facebook.com/plugins/comment_embed.php`
- `www.facebook.com/plugins/page.php`

If you want to test them all at once, you an use the [test page I created last week](https://fmarier.github.io/brave-testing/social-widgets.html).

fmarier:
Should we also move [these exceptions](https://github.com/brave/tracking-protection/blame/132ac934be9dd0113a4bb4d041f7602a72c60662/data/disconnect.json#L9348-L9375) to `brave-unbreak.txt` (if they're even needed at all)?

brave-unbreak.txt
```diff
@@ -88,3 +87,36 @@
...
+@@||facebook.com/connect/
+@@||www.facebook.com/connect
+@@||staticxx.facebook.com/connect/
+@@||graph.facebook.com/
+! Block fbcdn.net everywhere but allow Facebook embeds
+||fbcdn.net$third-party,domain=~facebook.com
+@@||staticxx.facebook.com/
+@@||xx.fbcdn.net/
+@@||www.facebook.com/*/plugin
+@@||www.facebook.com/plugins/
```
fmarier: If we unblock all of the plugins, then we'll let through the Like/Share buttons that @tomlowenthal wanted us to block in brave/brave-browser#3241:

- `www.facebook.com/plugins/like.php`
- `www.facebook.com/plugins/share_button.php`

fmarier:
Looping in @tomlowenthal because we may not want all of the widgets from <https://fmarier.github.io/brave-testing/social-widgets.html> to be allowed by default.

I think we definitely want Facebook videos to work for example, but maybe the LinkedIn Share/Follow social buttons aren't valuable enough to users warrant the tracking by Microsoft.

bbondy:
@fmarier @tomlowenthal I don't want to have that discussion here since this task is scoped to be about moving things from hardcoded lists to a unified exception list that we can build off of.  I'll remove the LinkedIn one and do it in another PR that we can only merge once the option to control it exists.

bbondy:
Mainly I suggest we avoid such discussions because if we just follow the 3 step plan I outlined in the section "Summary of the plan" from https://github.com/brave/brave-browser/issues/3475 then we can completely skip the discussions which have no clear way to decide what's best globally.

bbondy:
Talked to @fmarier on Slack and he's ok merging this since it retains the existing functionality.  Going ahead.

PR #53, opened 2019-02-27
Remove old copy of uBlock's unbreak list

fmarier:
In https://github.com/brave/ad-block/pull/183 we started fetching the upstream list directly.


fmarier:
We merged the fix to use the upstream list a week ago, I imagine we can now remove the old list safely.

PR #54, opened 2019-03-07
Add tags to optional filters

bbondy:
This adds tags to the optional filters.
A tagged filter is only matched if the ad_block_client has called addTag with that tag value. 
We have preferences now in brave-core to add those tags.

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -89,46 +89,52 @@
 @@https://analytics.*/piwik.$~third-party
 ! Fix for Suicide Prevention Lifeline Chat (brave/brave-browser#3492)
 @@||geoplugin.net/javascript.gp$script,xmlhttprequest,domain=suicidepreventionlifeline.org
-! Note that options will be added to exclude these filters soon. They
-! are added both as a blocking rule and as an exception rule so that
-! an exception is hit and will override what's in tracking protection protection.
-! Facebook logins
-||connect.facebook.com/*/sdk.js$script
-@@||connect.facebook.com/*/sdk.js$script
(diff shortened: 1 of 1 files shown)


brave-unbreak.txt
```diff
@@ -89,46 +89,53 @@
...
-||y8.com/js/sdkloader/outstream.js$third-party
+||platform.twitter.com/$tag=twitter-embeds
+@@||platform.twitter.com/$tag=twitter-embeds
+||syndication.twitter.com$tag=twitter-embeds
+@@||syndication.twitter.com$tag=twitter-embeds
+||pbs.twimg.com/$tag=twitter-embeds
+@@||pbs.twimg.com/$tag=twitter-embeds
+||cdn.syndication.twimg.com/$tag=twitter-embeds
+@@||cdn.syndication.twimg.com/$tag=twitter-embeds
+! LinkedIn in embed
```
bbondy: These LinkedIn tagged filters are new.

brave-unbreak.txt
```diff
@@ -89,46 +89,53 @@
 @@https://analytics.*/piwik.$~third-party
 ! Fix for Suicide Prevention Lifeline Chat (brave/brave-browser#3492)
 @@||geoplugin.net/javascript.gp$script,xmlhttprequest,domain=suicidepreventionlifeline.org
+! Block additional trackers
+||sp1.nypost.com$third-party
+||sp.nasdaq.com$third-party
+||lesechos.fr/xtcore.js$third-party
+||y8.com/js/sdkloader/outstream.js$third-party
```
bbondy: This whole block is just moved from the bottom above the tagged optional filters.

PR #61, opened 2019-05-29
Fix video playing on https://www.nbclosangeles.com

pes10k:
This is a tracking script, so only merge if webcompat > privacy


ryanbr:
Video was still broken, investigating a fix.  But this patch won't help (site has since changed, and whitelist no longer applicable.)

https://github.com/easylist/easylist/commit/63ce6f46a370f0827aef280d91fdcccf15367028

PR #62, opened 2019-06-05
Fixs Blank page brave/brave-browser#4049

ryanbr:
Stopped page blanking caused by dianomi.


ryanbr:
Done, made more specific to the inc.com domain

PR #66, opened 2019-06-07
Fix facebook logins on messenger.com brave/brave-browser#4173

ryanbr:
Allow facebook logins on messenger.com (its a facebook company afterall)


snyderp:
This makes sense to me, but let me make sure its a-ok with the other privacy folks invovled.  Will circle back ASAP

PR #67, opened 2019-06-07
Fix Merch direct powered sites brave/brave-browser#4752

ryanbr:
Generic due to the amount of sites using merch direct, also stripe.com is used on many sites.


ryanbr:
@snyderp something is stopping stripe.com from working properly, I basically went through various urls to reproduce the users issue. and js.stripe.com was only thing that did it.

ryanbr:
Adjusted to all cookies vs 3rd party will fix it instead. closing.

PR #68, opened 2019-06-07
Fix missing sigin icon on https://app.mysms.com/#register and https://app.mysms.com/#login

ryanbr:
Minor bug, but the google icon is blocked in error, and should be fixed

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -140,6 +140,7 @@
 ||video.twimg.com/$tag=twitter-embeds
 @@||video.twimg.com/$tag=twitter-embeds
 ! Fix sign in icon on https://app.mysms.com/#login
+||developers.google.com/identity/$image,domain=mysms.com
 @@||developers.google.com/identity/$image,domain=mysms.com
 ! Fix page rendering due to issues with scorecardresaerch.com https://github.com/brave/brave-browser/issues/1580
 ||scorecardresearch.com^$domain=dailymotion.com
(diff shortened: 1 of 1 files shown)


brave-unbreak.txt
```diff
@@ -139,6 +139,8 @@
 @@||api.twitter.com/1.1/$tag=twitter-embeds
 ||video.twimg.com/$tag=twitter-embeds
 @@||video.twimg.com/$tag=twitter-embeds
+! Fix sign in icon on https://app.mysms.com/#login
+@@||developers.google.com/identity/$domain=mysms.com
```
snyderp: Could we limit this to images?  Seems like it'd more tightly solve the issue

snyderp:
I'm guessing this is being blocked in Brave b/c of brave/tracking-protection.  My understanding is that resources that match exception rules in filter lists bypass tracking protection (e.g. are allowed regardless of tracking-protection rules).  I'm verifying with @bbondy.  If so, this should be good to merge

bbondy:
If it is because of tracking protection then this alone wouldn't solve it. You'd need

```
||developers.google.com/identity/$domain=mysms.com
@@||developers.google.com/identity/$domain=mysms.com
```


Basically to know you'd look at this list:
https://github.com/brave/tracking-protection/blob/master/data/disconnect.json

If you find `developers.google.com` on any 3p list, then it is only allowed on those 3p lists. 
Which does seem to be the case, it is under `Google`. 
So I think you'd need the double rule that I mentioned.

PR #72, opened 2019-06-10
Anti-adblock on nuvo.net

ryanbr:
Admiral anti-adblock makes this site unusable if you have sheilds on.


snyderp:
@ryanbr I'm not seeing any anti-adblock messages on nuvo.net.  How can i trigger the anti-adblock, so I can check the anti-anti-adblock?

ryanbr:
Ah correction, will only apply when cosmetic filtering is being applied. Admiral is a Anti-adblock service which will be the first thing people will see once cosmetic filtering is patched. see;

```
https://www.politicususa.com/
https://www.al.com/ 
https://www.pinknews.co.uk/
```

There is no damage deploying it earlier (just blocking their specific adserver domain), but can be delayed until cosmetic filtering is landed.

snyderp:
I'll create a tag to keep track of these things (things that it makes sense to merge in once we do cosmetic filters)

ryanbr:
ad domain no longer being used, closing this

PR #79, opened 2019-06-12
Fix video playback on engadget.com

ryanbr:
Video won't playback under "Featured Video", this should fix it.

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -143,6 +143,8 @@
 ||scorecardresearch.com^$domain=dailymotion.com
 ! https://community.brave.com/t/ad-not-bloked-properly/63628
 ||readthedocs.org/api/v2/sustainability/$script,domain=pyexcel.org
+! Embedded video on engadget.com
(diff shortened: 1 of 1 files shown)


ryanbr:
Not sure how this approach;  since each video is using a specific checksum
```
https://delivery.vidible.tv/jsonp/pid=576b1dbfe4b0c3d3818a1a7b/vid=5d0023a7e98ab543c5cd4237/564f3144ff690c0a7c285e51.js
https://delivery.vidible.tv/jsonp/pid=56f2c711e4b0419133b37121/vid=5cfedaebb8bad324d688e90d/564f3144ff690c0a7c285e51.js
```

snyderp:
I believe it doesn't need to be an exact match, it just needs to both hit a rule and an exception rule to bypass tracking protection. 

So, something like
```
||delivery.vidible.tv/jsonp/pid=$script,domain=engadget.com
@@||delivery.vidible.tv/jsonp/pid=$script,domain=engadget.com
```

Should do the trick.  @bbondy can you confirm, just to make sure we've got it 100% going forward?

PR #82, opened 2019-06-13
Possible fix for brave/brave-browser#4201

ryanbr:
A number of german sites, will check for multiple urls to load. This will be as base line commit, may require other changes if it doesn't work.

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -175,7 +175,6 @@
 ! spiegel.de (https://github.com/brave/brave-browser/issues/4201)
 ||imagesrv.adition.com/banners/$image,domain=spiegel.de
 ||googletagservices.com/tag/js/gpt.js$script,domain=spiegel.de
-||mxcdn.net/bb-mx/$script,domain=spiegel.de
 ||doubleclick.net/ddm/$image,domain=spiegel.de
 @@||doubleclick.net/ddm/$image,domain=spiegel.de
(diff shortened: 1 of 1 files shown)


ryanbr:
Its quite a complex Anti-adblock tech, so I'm not entirely sure if this will stop it. but it's a start.

brave-unbreak.txt
```diff
@@ -172,6 +172,15 @@
...
 ! Anti-adblock: dreamdth.com
 @@||dreamdth.com/js/wutime_adblock/ads.js$script
+! spiegel.de (https://github.com/brave/brave-browser/issues/4201)
+||imagesrv.adition.com^$image,domain=spiegel.de
+||googletagservices.com/tag/js/gpt.js$script,domain=spiegel.de
+||mxcdn.net^$script,domain=spiegel.de
+||research.de.com^$script,domain=spiegel.de
+@@||googletagservices.com/tag/js/gpt.js$script,domain=spiegel.de
+@@||mxcdn.net^$script,domain=spiegel.de
+@@||research.de.com^$script,domain=spiegel.de
```
snyderp: This still seems over broad.  There is no reason to do the block / except pattern except when
1) we want the resource to be loaded
2) its being blocked by tracking protection (i.e. the domain appears in this list: https://github.com/brave/tracking-protection/blob/master/data/disconnect.json)

ryanbr:
all these filters are needed to prevent ads/anti-adblock, the domain checks may change in the future. but this works. Similar filters incoming for golem.de

PR #83, opened 2019-06-13
Fixes brave/brave-browser#3418

ryanbr:
Considered anti-adblock because they're always countering whitelists/checking Easylist filters.


ryanbr:
Looking back, I've decided to merge into Easyprivacy since its directly related to this.

https://github.com/easylist/easylist/commit/357e3ee4898f43b3036af2b2a041f18d11ffcf13

PR #84, opened 2019-06-13
split brave rules across more descriptive files, fixes #81 [NO NOT MERGE]

### LINKED ISSUE #81: Create a new `brave-ads.txt` list
Now that we have [entries in `brave-unbreak.txt`](https://github.com/brave/adblock-lists/blob/b67bfb158a0862a83fd8370cdc3d9abb818ddf31/brave-unbreak.txt#L149-L150) which don't related to unbreaking sites for webcompat reasons, we should have a second list whose name better reflect the purpose of these entries (e.g. `brave-ads.txt` or just `brave.txt`).

That way we're transparent about what we're doing and we're not hiding extra blocking rules in a webcompat file.

pes10k:
This should only be merged once corresponding changes are made to ~https://github.com/brave/ad-block/blob/master/lists/default.h~ https://github.com/brave/adblock-resources/blob/master/filter_lists/default.json


ShivanKaul:
This is pretty stale at this point, closing, we should still do this type of cleanup though.

PR #86, opened 2019-06-15
Anti-adblock on animefever.tv (prevents video playback)

ryanbr:
Filter exists in Easylist, used to detect adblock to disable playback.

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -175,6 +175,8 @@
 @@||wallpapershome.com/scripts/ads.js$script
 ! Anti-adblock: mediaite.com
 @@||mediaite.com/adbait/adsbygoogle.js
+! Anti-adblock: animefever.tv
+@@||imasdk.googleapis.com/js/sdkloader/ima3.js$xmlhttprequest,domain=animefever.tv
 ! Anti-adblock: dreamdth.com
 @@||dreamdth.com/js/wutime_adblock/ads.js$script
 ! spiegel.de (https://github.com/brave/brave-browser/issues/4201)


snyderp:
This is an interesting case.  The code is looking for the error code of requesting `https://imasdk.googleapis.com/js/sdkloader/ima3.js`, not the actual contents.  But here they're just doing a `HEAD` request, so something funny is up.  

We still need to be careful to not loading things from google or other resources that will cause privacy problems.  So, in general, we should white list things that will unbreak sites, but only if it wont harm privacy.  If we have to choose between unbreaking or privacy, lets raise the conversation to figure a third option :)

okiehsch:
If you support `redirections`
````
||imasdk.googleapis.com/js/sdkloader/ima3.js$xmlhttprequest,redirect=nooptext,domain=animefever.tv
````
will fix the issue.

**Comment by snyderp** [comment]
@okiehsch yep!  Brave basically `redirect=nooptext` for all blocks (return empty 200 for all blocked resources) which helps in a lot of situations.

The tricky bit here is that its a HEAD request, which is doing something tricky here in or blocking system.  Digging more :)

Relevant code from the site
```
adsBlocked: function(t) {
    var e = new Request("https://imasdk.googleapis.com/js/sdkloader/ima3.js",{
        method: "HEAD",
        mode: "no-cors"
    });
    fetch(e).then(function(e) {
        t(!1)
    }).catch(function(e) {
        t(!0)
    })
},
```

okiehsch:
> Brave basically redirect=nooptext for all blocks (return empty 200 for all blocked resources) which helps in a lot of situations.

Hm, well I can reproduce the issue with uBO using Chromium and Firefox and 
````
||imasdk.googleapis.com/js/sdkloader/ima3.js$xmlhttprequest,redirect=nooptext,domain=animefever.tv
`````
works on my end, so it should work if one uses Brave then no?

snyderp:
right, i think the issue is that we're not handling the HEAD request as if it was `redirect=nooptext` using our own adblock lib.  Digging further :)

ryanbr:
I'm fine with this.

ryanbr:
Fixed in upstream uBO. Closing

Issue #65, opened 2019-06-06
[tracking] Fixes that should be reversed when uBO resources/polyfils are supported


snyderp:
https://github.com/brave/adblock-lists/pull/64

ShivanKaul:
@ryanbr can you make sure we do this cleanup?

ryanbr:
Really old, we can close this

PR #87, opened 2019-06-15
Anti-adblock on techsolutiontips.com

ryanbr:
Site checks on 2 adserver scripts to load, filter exists in Easylist. Will show up after x-mins and will keep nagging the end user.

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -169,6 +169,9 @@
 ! Fix facebook logins on messenger.com https://github.com/brave/brave-browser/issues/4173
 @@||facebook.com/login/$domain=messenger.com
 @@||connect.facebook.net^$domain=messenger.com
+! Anti-adblock: techsolutiontips.com
+@@||googlesyndication.com/pagead/js/r20190612/r20190131/show_ads_impl.js$domain=techsolutiontips.com
+@@||googlesyndication.com/pagead/js/adsbygoogle.js$domain=techsolutiontips.com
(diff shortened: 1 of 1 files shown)


brave-unbreak.txt
```diff
@@ -169,6 +169,9 @@
 ! Fix facebook logins on messenger.com https://github.com/brave/brave-browser/issues/4173
 @@||facebook.com/login/$domain=messenger.com
 @@||connect.facebook.net^$domain=messenger.com
+! Anti-adblock: techsolutiontips.com
+@@||googlesyndication.com/pagead/js/r20190612/r20190131/show_ads_impl.js$domain=techsolutiontips.com
```
snyderp: This rule looks specific to the date you visited the page on.  Needs to be generalized

snyderp:
This looks like the same issue from https://github.com/uBlockOrigin/uAssets/issues/3880

Since it'd be good to avoid any google ads loading on any pages in Brave, I think the best thing to do is to try and raise with the uBO folks, so that we can get this fix for free, w/o the privacy loss, when https://github.com/brave/adblock-lists/issues/65 lands.

@ryanbr if that makes sense to you too, lets follow your suggestion of upstreaming the issue to uBO and waiting for that to land in Brave (especially if this isn't a brave-reported issue). 

If that sounds good, can you close the issue?

snyderp:
https://github.com/uBlockOrigin/uAssets/commit/b6d4ac06694458af00a160b2ee3a00035f2293d6

Fixed in UBO

ryanbr:
Closing, fixed in uBO. safe to close.

PR #93, opened 2019-06-24
Fix Foxnews and Foxbusiness video playback

ryanbr:
Video playback was broken, this will fix it. Existing in Easyprivacy, but Disconnect was overruling it.

Included a Tracking /ads.js whitelist to track users with.

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -190,7 +190,6 @@
 @@||mxcdn.net/bb-mx/$script,domain=spiegel.de
 @@||imagesrv.adition.com/banners/$image,domain=spiegel.de
 ! Fix foxnews video playback
-||fncstatic.com/static/isa/app/lib/VisitorAPI.js$script,domain=foxbusiness.com|foxnews.com
 @@||fncstatic.com/static/isa/app/lib/VisitorAPI.js$script,domain=foxbusiness.com|foxnews.com
 ! Adblock-Tracking: foxnews.com / foxbusiness.com
 @@||fncstatic.com/static/v/all/js/ads.js$script,domain=foxbusiness.com|foxnews.com
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -189,6 +189,11 @@
 @@||googletagservices.com/tag/js/gpt.js$script,domain=spiegel.de
 @@||mxcdn.net/bb-mx/$script,domain=spiegel.de
 @@||imagesrv.adition.com/banners/$image,domain=spiegel.de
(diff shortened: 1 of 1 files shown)


ryanbr:
Correction, block not needed.

brave-unbreak.txt
```diff
@@ -189,6 +189,10 @@
 @@||googletagservices.com/tag/js/gpt.js$script,domain=spiegel.de
 @@||mxcdn.net/bb-mx/$script,domain=spiegel.de
 @@||imagesrv.adition.com/banners/$image,domain=spiegel.de
+! Fix foxnews video playback
+@@||fncstatic.com/static/isa/app/lib/VisitorAPI.js$script,domain=foxbusiness.com|foxnews.com
+! Adblock-Tracking: foxnews.com / foxbusiness.com
+@@||fncstatic.com/static/v/all/js/ads.js$script,domain=foxbusiness.com|foxnews.com
```
bbondy: If disconnect is breaking this ONLY, then you need to make the adblock lib search for an exception. To do that.
So I think you need to duplicate these rules with both block and exceptions.

! Fix foxnews video playback
||fncstatic.com/static/isa/app/lib/VisitorAPI.js$script,domain=foxbusiness.com|foxnews.com
@@||fncstatic.com/static/isa/app/lib/VisitorAPI.js$script,domain=foxbusiness.com|foxnews.com
! Adblock-Tracking: foxnews.com / foxbusiness.com
||fncstatic.com/static/v/all/js/ads.js$script,domain=foxbusiness.com|foxnews.com
@@||fncstatic.com/static/v/all/js/ads.js$script,domain=foxbusiness.com|foxnews.com

snyderp: @bbondy I don't see fncstatic.com in the tracking protection data

https://github.com/brave/tracking-protection/blob/master/data/disconnect.json

My understanding was that if the domains not in there, the tracking protection is def not involved.  Is that incorrect?

ryanbr: Caused by;
```
easyprivacy_general.txt:/VisitorAPI.js
```
And whiteslist in EP:
```
easyprivacy_whitelist.txt:@@||fncstatic.com^*/visitorapi.js$script,domain=foxbusiness.com|foxnews.com
```

snyderp: Yes, only referring to the block and then allow pattern.  That pattern is specifically only needed when working around tracking protection.

ryanbr: So is it safe to merge? no blocking filter needed because its specific to EP only and not the disconnect list

PR #98, opened 2019-06-27
Fix another video bug on Foxnews / Foxbusiness

ryanbr:
Another bug reported, since it exists in Easylist already, but because the Disconnect list is overruling it. I wasn't aware it was still an issue.  This should hopefully fix it, hopefully.


snyderp:
Is this needed in a post-uBO world?

ryanbr:
Will still be needed, related more to the disconnect list than ubo

snyderp:
I see this as blocked in uBO though, any guess on why we'd to except-rule it, and uBO doesn't.  If its related to tracking protection, its some other resource (since, this one is blocked in uBO).

PR #99, opened 2019-06-28
Fix https://deora.dev/corgi/ and also fix any other sites using Reddit api

ryanbr:
2 changes, one for Brave Nightly (which has the wildcard fix) and fix the site in question.

Was notified via twitter; https://twitter.com/BrendanEich/status/1144377788427730945

We can remove the deora.dev once adblock-rust is in release.

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -83,6 +83,9 @@
 @@||gateway.reddit.com^
 ! https://github.com/brave/adblock-lists/issues/39
 @@||alb.reddit.com^
+! Allow sites to use reddit api
+||reddit.com^*/.json?$script,third-party
+@@||reddit.com^*/.json?$script,third-party
 ! DDG 1P analytics and optimization
 @@||improving.duckduckgo.com^$~third-party
 ! Disable PDFJS which we include by default's telemetry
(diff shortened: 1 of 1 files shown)


snyderp:
Re https://twitter.com/fanboynz/status/1144432292749701120

Do not comment on issues publicly before we've actually merged a fix in.  That puts Brave in a very bad spot if the fix doesn't work, needs to be tested, etc

snyderp:
I'm worried that allowing the reddit apis will allow tracking to occur.  They are listed in disconnect as a tracking party, for example, and the change made is very broad.  Can you confirm that the *.json rules are not tracking related? @ryanbr 

Basically, if allowing more resources would enable users to be tracked, we need to come up with a different solution (unless its a very very short term solution).

ryanbr:
I've look through the reddit docs it seemed pretty sane?  Made an assumption it would be a non-issue fix, my error. The third-party json is just referencing reddit functions directly. I can't say definitely that reddit.com won't track json requests, most likely reddit will be tracking how many times/how popular a api is being hit from user oauth id# and maybe any referrer clickthroughs.

The actual json for corgi (cleaned up), didn't really show anything noticeable.
`https://paste.ee/p/LHEnY`

`https://stackoverflow.com/questions/6183318/getting-new-posts-from-a-subreddit-in-json`
`https://alpscode.com/blog/scripting-with-reddit-api/ `
`https://googlechrome.github.io/samples/fetch-api/fetch-reddit.html`
`https://www.reddit.com/dev/api`

Reddit was tracking users directly on the main site (which I was involved with);  `https://www.reddit.com/r/technology/comments/87s1fl/psa_reddit_has_enhanced_their_tracking_they_now/`

Also are blocking reddit tracking pixels already (see the GET response details below);

`https://www.reddit.com/r/redditdev/comments/3xm9x9/how_do_i_add_the_raw_json_parameter/`
`||pixel.redditmedia.com^`

snyderp:
Verified that the request for json file does not carry identifiers, cookies, referrers, etc

PR #100, opened 2019-06-29
Video playback is broken on https://www.zdnet.com

ryanbr:
Made it domain specific, but it may need to be generic depending on how popular this specific script is.

Filter is blocked via disconnect; this fix's it.

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -164,8 +164,8 @@
 ! Fix addthis.com issues on rhmodern.com  https://github.com/brave/brave-browser/issues/3653
 @@||s7.addthis.com^$script,domain=rhmodern.com
 ! 2mdn video playback script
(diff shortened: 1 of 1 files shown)


ryanbr:
Test url;

`https://www.zdnet.com/article/brave-defies-googles-moves-to-cripple-ad-blocking-with-new-69x-faster-rust-engine/`

snyderp:
This is another "the fix enables tracking" fix (might not be any other way), but want to get an official policy set internally before deciding on these.  Blocking for the moment on the #ads-blocking convo

BrendanEich:
@snyderp We have the policy worked out, can you put it in this repo's wiki so we can all refer to it? Thanks.

snyderp:
@BrendanEich done and done
https://github.com/brave/adblock-lists/wiki/Policy-on-Filter-List-Exceptions

PR #101, opened 2019-06-29
Fix video playback on https://www.maxpreps.com/

ryanbr:
Sample video; 

`https://www.maxpreps.com/video/watch/new-jersey-s-most-dominant-football-teams?v=36c936cb-5ffe-4dbe-8688-521ab2300c23`

Similar issue to the Foxnews issue, same script is used; imasdk.googleapis.com/js/sdkloader/ima3.js

Filter exists in Easylist. Disconnect list overrules the Easylist_whitelist we have.


snyderp:
Pausing on this for the same reason as https://github.com/brave/adblock-lists/pull/100.  Tagged accordingly so that we can address these all at the same time once policy / decision is made

PR #103, opened 2019-06-30
Fix images not displaying on drudgereport.com

ryanbr:
Fixed in Adblock-Rust, but not in Brave release/beta. Just quick fix to for images being blocked. Can be removed once Adblock-Rust lands in Release

Reported here; https://old.reddit.com/r/brave_browser/comments/c7f44y/brave_blocks_pictures_from_drudgereport_if/


ryanbr:
Interesting, it seemed to fix itself.

ryanbr:
Okay it's still being reported, should be safe to merge. Possibly revolving image servers. Twimg is Twitter, so they're pulling images from them..

ryanbr:
Okay its safe, I suspect as much; https://old.reddit.com/r/brave_browser/comments/c7f44y/brave_blocks_pictures_from_drudgereport_if/esh70dp/?context=3

PR #106, opened 2019-07-02
saturn exception from tracking protection

pes10k:
So that we don't need a special case in tracking protection logic


fmarier:
Is `google-analytics.com` blocked globally in EasyList?

snyderp:
it is, but if there is an exception rule, it overrides…

ryanbr:
If no one has issues, It's an easy fix in Easyprivacy. , The 3 whitelists (I prefer not blank rule these filters)

`@@||googletagmanager.com/gtm.js$script,domain=saturn.at`
`@@||google-analytics.com/analytics.js$script,domain=saturn.at`
`@@||google-analytics.com/plugins/ua/ec.js$script,domain=saturn.at`

ryanbr:
Seems saturn.at is listed;
https://github.com/brave/tracking-protection/blob/master/data/disconnect.json#L10762 

But googletagmanager.com seems to cause a few issues. It'll stop google-analytics from loading.

`||googletagmanager.com/gtm.js$script,domain=saturn.at`
`@@||googletagmanager.com/gtm.js$script,domain=saturn.at`

snyderp:
@fmarier can we confirm if these are even still needed.  This is getting hairy… Does someone on the security team have an android device, or similar?

Also, it seems likely that uBO resources will automatically fix in a few weeks anyway, so maybe all the more reason to just remove all special casing for staturn.at all together.

Thoughts?

fmarier:
Yes, it's still needed:

1. I tried removing it in https://github.com/brave/tracking-protection/pull/35.
2. It caused https://github.com/brave/brave-browser/issues/4248 on my Android phone.
3. I reverted my PR in https://github.com/brave/tracking-protection/pull/37.

ryanbr:
This might be a better option if we need to allow these scripts;

`||googletagmanager.com/gtm.js$script,domain=saturn.at`
`@@||googletagmanager.com/gtm.js$script,domain=saturn.at`
`||google-analytics.com/analytics.js$script,domain=saturn.at`
`@@||google-analytics.com/analytics.js$script,domain=saturn.at`
`||google-analytics.com/plugins/ua/ec.js$script,domain=saturn.at`
`@@||google-analytics.com/plugins/ua/ec.js$script,domain=saturn.at`

Would that work better? I'd be more specific. and we can revert some of these filters when the ubo stuff is landed.

fmarier:
I'm not entirely sure we need to enable `googletagmanager.com` since the only exception on our copy of the Disconnect list is for `google-analytics.com`.

ryanbr:
> I'm not entirely sure we need to enable `googletagmanager.com` since the only exception on our copy of the Disconnect list is for `google-analytics.com`.

These are being loaded; (direct copy/pasta)

`https://www.googletagmanager.com/gtm.js?id=GTM-KLM5RV`
`https://pubads.g.doubleclick.net/gampad/adx?iu=/21792584450/Test/Testpixel&sz=1x1&t=testpixel%3DSE_Startseite_Mobile&c=[random]`
`https://service.maxymiser.net/api/eu/saturn.at/eb28b4/mmapi.js`

ryanbr:
This fine below (from https://github.com/brave/tracking-protection/blob/master/data/disconnect.json)

```
    ],
    "Brave": [
      {
        "Saturn": {
          "https://www.saturn.at/": [
            "google-analytics.com"
          ]
        }
      }
    ]
```
The main issue here is googletagmanager.com/gtm.js If this script is blocked from loading; it'll stop google-analytics from initiating (rendering the Saturn/disconnect.json pointless). If the disconnect list isn't operating as suggested, then further whitelists/blocklists (see https://github.com/brave/adblock-lists/pull/106#issuecomment-508298939) will be needed. But you will still need the **googletagmanager.com/gtm.js** fix regardless.   Some google scripts rely on other scripts to load correctly (may/maynot fix by the ubo changes)

I hope that clarify it. I'll implement a fix for Easyprivacy since it was easily reproducible.

fmarier:
The reason why I'm not sure that the GTM exception is needed is that we currently don't have it and the site loads fine on Android (with the GA exception you pasted in https://github.com/brave/adblock-lists/pull/106#issuecomment-508373299).

Did you examine the site on Android or on desktop? Because it is different in some way. I was unable to reproduce the breakage until I visited the site on my Android phone.

ryanbr:
> Did you examine the site on Android or on desktop? Because it is different in some way. I was unable to reproduce the breakage until I visited the site on my Android phone.

2 ways I tested. 

In a private desktop window with a mobile UA (which replicated the site), Using Brave.
On Android Firefox and ubo to show blocked items.

Both showed googletagmanager.com. This might be fixed in ubo filters (disabled ubo specific filters for testing), but without it you'll need to whitelist this. Can you confirm using either of my test cases?

fmarier:
I didn't have much luck [testing the mobile site on desktop](https://github.com/brave/tracking-protection/pull/35#discussion_r278142919) when I last looked at this, so I'm not sure we can rely on that first test case.

As for uBlock Origin in Firefox for Android, that sounds like a good test, but it's a different enough environment that I'm not sure we can rely on it either.

The only way I can think of to test this is to:

1. Push out the change in `brave-unbreak.txt`
2. Remove the `saturn.at` customization from our copy of the Disconnect list.
3. Test in Brave for Android.
4. Hotfix `brave-unbreak.txt` or restore the Disconnect customizations if *step 3* fails.

ryanbr:
We can close this, we have whitelists to fix this site in Easyprivacy. 

`@@||google-analytics.com/analytics.js$domain=saturn.at|schweizerfleisch.ch`
`@@||googletagmanager.com/gtm.js?$domain=bunte.de|finanzen.net|saturn.at`
`@@||google-analytics.com/plugins/ua/ec.js$domain=saturn.at`

Re-open this ticket if we're having issues still, but seems safe to close based on the whitelists.

PR #107, opened 2019-07-03
Fix Anti-adblock on vox (theverge.com etc) sites

ryanbr:
Just a WIP. Not sure if this is the only dodgy script or not. 

Seems these 2 concert.io domains are being generated in Brave but not in ubo.
```
https://cdn.concert.io/lib/adblock/verge-01-desktop.png
https://cdn.concert.io/lib/adblock/verge-01-mobile.png
```


ryanbr:
Okay check the script (required a few clear cookies/cache) but this script seems to load the additional elements causing some page delays.

snyderp:
@ryanbr looks like this PR has become stale.  Can you fix the conflicts and re-submit?

ryanbr:
Just a quick adjustment in GH.

PR #108, opened 2019-07-04
Allow first party on maxmind.com  https://community.brave.com/t/stop-blocking-the-website-for-maxmind-com/68326

ryanbr:


snyderp:
Merging these in, but they'll no longer be needed when we do the disconnect -> adblock change, so tagging revist

PR #113, opened 2019-07-05
Fix sites using embeded video from vidible.tv

ryanbr:
Messed the last commit, re-up'd a new pull.  A few sites are broken due to vidible.tv being blocked via disconnect. This will modify our current filter to be generic to allow videos.

Sample videos: 
`https://popculture.com/tv-shows/2019/07/03/seinfeld-leaving-hulu-migrate-warnermedia-streaming-service/`
`https://comicbook.com/2019/07/04/mad-magazine-shutdown-details-revealed/`
`https://www.stuff.co.nz/entertainment/celebrities/114009167/joe-jonas-and-sophie-turner-share-their-first-wedding-photo`

This script loads the video, and needs to be allowed:
`https://delivery.vidible.tv/jsonp/pid=58d1e0f7f78ced6518eee162/vid=5d19ece5b8bad37cea19c3a3/58d1b4a95095491a900c189d.js?.embeded=cms_video_plugin_chromeExtension&m.sitesection=world&m.disableadspreloading=true`

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -154,9 +154,9 @@
 @@||captcha.vresp.com^$domain=lawfirmkpi.com
 ! https://community.brave.com/t/ad-not-bloked-properly/63628
 ||readthedocs.org/api/v2/sustainability/$script,domain=pyexcel.org
-! Embedded video on engadget.com
-||delivery.vidible.tv/jsonp/$script,domain=engadget.com
-@@||delivery.vidible.tv/jsonp/$script,domain=engadget.com
+! Embedded vidible video's
+||delivery.vidible.tv/jsonp/$script
+@@||delivery.vidible.tv/jsonp/$script
 ! LinkedIn in embed
 ||platform.linkedin.com/$tag=linked-in-embeds
 @@||platform.linkedin.com/$tag=linked-in-embeds
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -155,8 +155,10 @@
 ! https://community.brave.com/t/ad-not-bloked-properly/63628
 ||readthedocs.org/api/v2/sustainability/$script,domain=pyexcel.org
 ! Embedded vidible video's
-@@||vidible.tv/prod/$script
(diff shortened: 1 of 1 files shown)


snyderp:
I feel nervous about this one, since its a web-wide change, and the included code does potentially privacy harming things (i see ajax calls included, for example, in the source).  

I think this one might fall in the "tag for future consideration" category.  Though, could you check and see if the network requests *made* by the scripts allowed by these changes would be blocked, then i think its safe to merge in. 

But if it'll result in additional network requests that are not blocked, lets stash this one in the "for future investigation" category

ryanbr:
The disconnect list shouldn't be blocking vidible.tv out right. vidible.tv has their own tracking domains, which is already blocked in EP.  This is just for usability, otherwise people will complain or switch to another browser because video playback is broken. Making domain specific filters won't work because vidible.tv is on so many sites. https://publicwww.com/websites/delivery.vidible.tv/

You can test these sites with sheilds=Off, and just enable ubo. video works fine,

snyderp:
When i add those rules to my brave adblock (brave://adblock) the videos still don't load. Looks like https://delivery.vidible.tv/jsonp/* is just the first part of a chain that needs to load.

I also added
```
||vidible.tv/prod/player/js$script
@@||vidible.tv/prod/player/js$script
```

to allow https://cdn-ssl.vidible.tv/prod/player/js/21.1.143/vidible-min.js to load, but also didn't get videos playing.  Can you double check these changes?  If I'm doing something wrong, can you let me know how reproduce the bug / fix?

ryanbr:
Ah yes I missed it because it was already included in Easylist.

`@@||vidible.tv/prod/$third-party,script`

Note; vidible.tv/prod/player/js$script was probably too restrictive. while it may work, it could possible have other issues we're not aware of. It'll be blocking legit content. 

Just noticed a suspect element; (blocked)

`https://cdn.vidible.tv/prod/player/ad-time/local-storage.html`

The latest patch show allow the video and the features while still blocking the trackers/suspect elements.
vidibletv

PR #116, opened 2019-07-05
Japanese specific filters

ryanbr:
Only 2 generics,  rest a site specific or third-party. only applicable for Japanese region.


ryanbr:
Not sure what label to give this, created a regional tag. Which will cover any regional requests.

snyderp:
@ryanbr thanks for these!  Some seem very very specific though.  Would it make sense to generalize the ones that are single-resource/file specific currently?

ryanbr:
Only reason some of these are specific, because it'll cause false positives with other elements used on the site. Most of these specific filters would be reverted once we get the cosmetic filtering.

PR #119, opened 2019-07-07
Fix blocking of aliexpress due to overreaching filter in ABP-Japanese

ryanbr:
We shouldn't need to create this filter like this, but it seems they have a history of blocking Aliexpress for some reason. It's a legit shopping site (which I've personally used also).

Logged an issue request; https://github.com/k2jp/abp-japanese-filters/issues/56 (which is being ignored)

https://github.com/k2jp/abp-japanese-filters/issues?utf8=✓&q=is%3Aissue+aliexpress

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -237,6 +237,8 @@
 ||static.maxmind.com^$~third-party
 @@||blog.maxmind.com^$~third-party
 @@||static.maxmind.com^$~third-party
+! ABP Japanese blocking Aliexpress (https://github.com/k2jp/abp-japanese-filters/issues?utf8=✓&q=is%3Aissue+aliexpress)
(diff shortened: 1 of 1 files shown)


snyderp:
@ryanbr can we limit this somehow, since there are also tracking resources on aliexpress?  Maybe only adding the exception for 3p resources, something like that?

ryanbr:
Would ||aliexpress.com^$badfilter work better?

ryanbr:
If ||aliexpress.com^$badfilter isn't an option. Here is a 2 screenshots of @@||aliexpress.com^$~third-party

Front page:
aliexpress-front

Item page:
aliexpress-item

It looks okay, All the trackers are 3rd-party rather being self-hosted, and all blocked in EP.

PR #120, opened 2019-07-07
Adblock Tracking Script on https://www.healthline.com/

ryanbr:
Just Adblock tracking script;

`https://www.healthline.com/scripts/advertising.js`

Script /advertising.js is blank.


snyderp:
@ryanbr what page can I trigger this on?  I wouldn't expect Brave (b/c we 200 block requests) would be detectable in this case.  Can you share a URL that has this issue?

ryanbr:
Can trigger it directly from the home / any page; 
healthline

snyderp:
Where can i trigger the "anti-adblock" behavior though?  I dont see it at either the landing page or child pages

ryanbr:
Its anti-dblock in the sense it's tracking adblock users, not anti-adblock where it prevents access. (similar to other adblock tracking blocks previously submitted)

snyderp:
ah, got it, thanks!

PR #123, opened 2019-07-09
Remove filter, causing issues on rakuten.co.jp

ryanbr:
Breaks site, lets remove it.


ryanbr:
Only affects Release.  breaks a few elements in error;

PR #125, opened 2019-07-11
Stop wp pl ads from being generated

ryanbr:
Intentionally didn't link directly to the github ticket due to potential referrer tracking to the site (trying to avoid any unwanted attention / arms-race)

Blocking this one script, stops ads and trackers from being generated and possibly making that site (and other wp.pl related sites) less buggy and quicker to load.

Tagged as regional (Polish sites only)

wps

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -254,6 +254,8 @@
 @@||reddit.com/comments/$domain=batcommunity.org
 ! thehindu.com (https://github.com/brave/brave-browser/issues/4808)
 @@||thgim.com/static/js/ads.min.js$script,domain=thehindu.com
+! stop wp.pl ads from generating (https://github.com/brave/brave-browser/issues/4914)
+||wpcdn.pl/wpjslib/$script,third-party
 ! Fix twitter images 
 ||twimg.com^$image,domain=drudgereport.com|batcommunity.org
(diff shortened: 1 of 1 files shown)


ryanbr:
Sample url;

`https://www.dobreprogramy.pl/AMD-tlumaczy-Procesory-Zen-2-sa-odporne-na-wszystkie-znane-luki-sprzetowe,News,102339.html`

snyderp:
@ryanbr thanks for digging into this.  Its important to tie the list addition to the GH issue though formally.  If you think the arms race stuff might get real heated, lets figure out another way to work around the problem.  Would be good to bring @rebron in if so to discuss, just to make sure he knows which items are fixed.

Thanks for getting this sorted, looks like a good win :)

ryanbr:
The main issue was the original GH ticket request had a direct link to the site, if it was edited (quickly) with a `url comment` instead. would avoid referrer trackbacks. We've had instances where there was arms race because the site was aware of the GH ticket.  

The same issue with reddit requests, where users publish the site urls directly. Makes it hard to comment  on fixes knowing the site would be now monitoring thread. We can't fix reddit requests, but we have some control on GH requests.

snyderp:
Whats the best way of tying the fix to the Brave GH issue then?  That part is necessary.  If you have a suggestion / preference on other ways of doing so though, I'm a-ok to have us move to it when necessary

PR #127, opened 2019-07-12
Fix ebay image upload brave/brave-browser#5190

ryanbr:
Easy fix, specific for ebay, made it generic because it could be affecting a few scripts used by ebay

The specifc script:

`https://msa-b1.cgi5.ebay.com/ws/eBayISAPI.dll`


ryanbr:
Also confirmed no trackers use this subdir. But can make specific if need be.

snyderp:
@ryanbr looks good, but dont forget the "webcompat" label

PR #128, opened 2019-07-13
Adblock Tracking script on https://www.mediaite.com/

ryanbr:
Just Adblock tracking script;

As seen on;

https://www.mediaite.com/

source:
(blank js check)


snyderp:
Can this be restricted to just mediate.com?

snyderp:
could this be limited to `domain=mediate` though?  Just because, with the previously discussed policy (in the wiki) its easier for Brave to whitelist domain specific items.

PR #129, opened 2019-07-15
Merge outbrain scripts from Fanboy Annoyances List

ryanbr:
Disconnect already blocks outbrain.com, this just covers the scripts the load or check on outbrain. And one domain not covered by disconnect


ryanbr:
Came about because of this; `https://themargins.substack.com/p/taboola-outbrain-and-the-chum-supply`

snyderp:
I'm fine to add these in, but just curious, why not pull them up into EL or EP?

ryanbr:
While I considered them ads, some people found them useful for some reason and complained it'll cause false positives when they use outbrain.com to host internal links. IMO, using outbrain even for selfhosted links is a privacy issue.

PR #131, opened 2019-07-16
Fix login issues with ssrn.com due to use of adobedtm.com

ryanbr:
This should fix login issues on ssrn.com (we already have a adobedtm.com in EP for these scripts)


ryanbr:
can be tested on https://hq.ssrn.com/login/pubsigninjoin.cfm

PR #132, opened 2019-07-20
Fix Anti-adblock, forced ads on gfycat.com (reported https://community.brave.com/t/pages-loads-3rd-party-ads-after-page-loads/71472/10)

ryanbr:
As a user reported an issue with force ads due to use of 'Instart Logic', this can be reviewed once Ubo filter fixes land. but this commit should help mitigate most of it.


ryanbr:
Actually can't confirm its Instart logic ads. but the filter is valid still.

PR #137, opened 2019-07-25
Anti-adblock Tracking for https://animeflv.net/

ryanbr:
Ad block tracking:

`https://animeflv.net/ver/52455/dumbbell-nan-kilo-moteru-4`

Script in question:
`https://s3.animeflv.com/js/adsbygoogle.js`
Source of file:
`var loadingAds=true;`

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -208,6 +208,8 @@
 @@||wallpapershome.com/scripts/ads.js$script
 ! Anti-adblock: haaretz.com
 @@||haaretz.com/htz/js/advertisement.js
+! animeflv.com
+@@||animeflv.com/js/adsbygoogle.js$script,domain=animeflv.com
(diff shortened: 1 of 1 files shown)


brave-unbreak.txt
```diff
@@ -208,8 +208,10 @@
 @@||wallpapershome.com/scripts/ads.js$script
 ! Anti-adblock: haaretz.com
 @@||haaretz.com/htz/js/advertisement.js
-! animeflv.com
+! animeflv
+@@||animeflv.net/js/adsbygoogle.js$script,domain=animeflv.net
 @@||animeflv.com/js/adsbygoogle.js$script,domain=animeflv.com
+|animeflv.net
```
snyderp: Just double checking, is this correct or a typo?  According to https://adblockplus.org/filter-cheatsheet#blocking2, it looks like this needs to include to protocol to be useful, no?

ryanbr: My err. was a typo.

PR #139, opened 2019-07-25
Anti-adblock on thehindu.com

ryanbr:
Just a quick Anti-adblock script.

Seen on home page: `https://www.thehindu.com/`

`https://th.thgim.com/static/js/ads.min.js`
`Adblock=false;`


snyderp:
@ryanbr happy to approve, but do you expect uBO parity would change the need for this?

PR #140, opened 2019-07-25
Make gfycat.com block a little more specific

ryanbr:
||ga.gfycat.com^ might be too generic causing false positives.

Currently `https://gfycat.com/` is boken, this hopefully will help. if not I'll have another idea to fix this.

Blocks this script specifically, doesn't seem to change.
`https://daf77c42.ga.gfycat.com/3df94ccab771b057d2b03ff8c56ad826`

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -277,7 +277,7 @@
 ! stop wp.pl ads from generating (https://github.com/brave/brave-browser/issues/4914)
 ||wpcdn.pl/wpjslib/$script,third-party
 ! gifycat.com ads (https://community.brave.com/t/pages-loads-3rd-party-ads-after-page-loads/71472/10)
-||ga.gfycat.com^$script,domain=gfycat.com
+||daf77c42.ga.gfycat.com^$script,domain=gfycat.com
 ! Fix twitter images 
 ||twimg.com^$image,domain=drudgereport.com|batcommunity.org
 @@||twimg.com^$image,domain=drudgereport.com|batcommunity.org


snyderp:
> Currently https://gfycat.com/ is boken, this hopefully will help. if not I'll have another idea to fix this.

Not to be flip, but it would be good to know for sure if this will fix things before pushing changes to users.  Can you please test further until you have a fix that you're sure will solve the problem, or at least seems a bit more certain :)

ryanbr:
The main issue is the UBO vs Brave, they act differently on site. 

The first script that loads;

`https://daf77c42.ga.gfycat.com/3df94ccab771b057d2b03ff8c56ad826`

This stop here with ubo. But under Brave we get more ga scripts which i probably causing more issues. Then the 2nd script;

`https://szirlmfs.ga.gfycat.com/pynf7jseyeymqdhmvbldvvtcuwalvgri`

.ga stuff is an Anti-adblock script and the differences between Brave and UBO displaying different results. Blocking the first loaded ga script its the most logical way to approach it. The option is either we remove the filter completely or use this filter, because as it stands gfycat is broken on Brave (showing a blank page).

All the ga. scripts on gfycat load other ga scripts, there is dependency issues here. So a generic ga. block while it worked initially, they changed the scripts to enforce some dependency between scripts changed and broke our block. When a site has multple dependent scripts being loaded, I'll back-track until find the url that is causing the issue (ie, the first script)

It'll be much easier once we get UBO snippets. 

Related Snippet; https://github.com/uBlockOrigin/uAssets/pull/5983/commits/c36625b7ef133609f1e4d631924bf6efa9439771

snyderp:
Have you tried Brave://Adblock? You can add your rules there and see, in Brave, if it fixes the problem.  My loop is general diagnose the problem in uBO and then try the fixes directly in Brave. Makes it easier to know 100% if the fix works

ryanbr:
I haven't really got Brave://adblock to ever work reliably. even a whitelist of @@||ga.gfycat.com^  It'll still use the existing filter.

snyderp:
Yikes.  Okay, we need to address this ASAP.  I did not realize you were not testing rules in brave before shipping / merging.  We need all brave changes tested in brave before shipping / merging.

Can you give me an example of `brave://adblock` not applying a rule you expect, on a specific page?  I've had good luck w/ it since moving to the new rust version.  I can then make sure thats fixed ASAP.

ryanbr:
Site has since changed, no longer needed

PR #151, opened 2019-08-03
Stop Instart logic ad scripts on ign.com

ryanbr:
Instart logic adblock scripts are being used on ign.com. Was also committed here; https://github.com/abp-filters/abp-filters-anti-cv/pull/242  

Blocking sample script;
`https://ozqurewm.g01.ign.com/hylu0nfptvjbdquusdzrpiamwtvarryb`

On `https://au.ign.com/`


ryanbr:
Can add a $domain= if need be, but only specific for the subdomain on ign.com.

snyderp:
Yep, that'd be great @ryanbr; if you can add domain= rules for ign.com and any other needed domains.  I'll merge in then

PR #163, opened 2019-08-14
Fix blank page on https://bandai-hobby.net/sw/

ryanbr:
Blank page caused by maxmind on `https://bandai-hobby.net/sw/`

Fixed in EP: https://github.com/easylist/easylist/commit/86fe778758d8ee8f4814e23b2d1d1ab18c6ef736

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -298,6 +298,11 @@
 ||static.maxmind.com^$~third-party
 @@||blog.maxmind.com^$~third-party
 @@||static.maxmind.com^$~third-party
+! bandai-hobby.net (maxmind check causing blank pages)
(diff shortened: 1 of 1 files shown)


snyderp:
@ryanbr if this is fixed in EP, do we still need it here?

ryanbr:
maxmind is also in the disconnect list, would it be still valid?

snyderp:
Ah, i see, so won't be needed once https://github.com/brave/adblock-lists/blob/master/brave-disconnect.txt replaces disconnect/tracking protection in Brave?

ryanbr:
Probably a valid commit until that lands (was testing in Brave nightly) when I tested this commit

snyderp:
Sounds good.  I added another relevant tag (please add to other PRs if you can think of where it'd apply) so that we can pull this out when its no longer needed

PR #164, opened 2019-08-14
Allow twitter api readahead

ryanbr:
We were notified on twitter by @HubertGAM, that `https://twitter.com/HubertGAM/status/1161776282587271175` we're breaking twitter user retreival api.

I haven't got a login to test, but going by the user screenshot it seems safe enough?

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -86,6 +86,9 @@
 ! Allow sites to use reddit api
 ||reddit.com^*/.json?$script,third-party
 @@||reddit.com^*/.json?$script,third-party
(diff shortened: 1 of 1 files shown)


snyderp:
I dont think this is tracking protection related.  Can it be tied to domain=twitter.com?  If so, i'll merge in now

ryanbr:
Nah, its 3rd-party only. so if site wants to grab @usernames via twitters api which buffer.com did, would be highly likely it would affect other 3rd-party sites that used this twitter function.

ryanbr:
Somewhat similar to the reddit api issues we had

PR #170, opened 2019-08-20
Anti-adblock Tracking on reddit.com

ryanbr:
Applicable and only seen on `www.reddit.com` and `old.reddit.com`. Desktop only, mobile reddit seems to be not applicable

`https://www.redditstatic.com/desktop2x/js/ads.js`

**Source:**

`var e=document.createElement('div');`
`e.id='a2ba06a4-a2ec-4182-b295-c15ffe5f1181';`
`e.style.display='none';`
`document.body.appendChild(e);`

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -87,7 +87,7 @@
 ||reddit.com^*/.json?$script,third-party
 @@||reddit.com^*/.json?$script,third-party
 ! Adblock Tracking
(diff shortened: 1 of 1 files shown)


snyderp:
Can we generalize a bit more here?  `desktop2x` likely could take a couple of different values.  Otherwise, LGTM :)

ryanbr:
I wouldn't suspect it'd change, but I made it into a regex instead.

PR #189, opened 2019-09-04
Remove Blocking filters

ryanbr:
Since the need for blocking filter aren't needed, Removed Blocking filters, the whitelists should still be needed, so left alone.


ryanbr:
We should do this in 2 parts, removing the blocked items (leaving the whitelists) which is the safest option. 
2nd part would be to start reviewing / updating any filters.

snyderp:
@ryanbr is this ready for merge?

ryanbr:
Yup safe to merge

PR #195, opened 2019-09-08
Fix for dailymotion.com filter not required

ryanbr:
Looking at review, and was reported it was still an issue. Seems scorecardresearch.com was causing the issue, but in opposite. a white-list was needed.

https://github.com/easylist/easylist/commit/c8677d3b3db75b83a0e43fe2206eca47fa1942fa


ryanbr:
Was re-reported here; https://community.brave.com/t/dailymotion-still-not-working/81321

PR #204, opened 2019-09-18
Fix instart ads on ign.com

ryanbr:
Was reported by @Brave-Matt the current filter is causing some intermittent css issues. Just an adjustment to fix this, the url may need to change again since Instart ads (anti-adblock script) scripts will revolve.

`https://bwtjhnrr.g02.ign.com/c0fa4ccab926a090751edfd8cf66d529`

As seen on `https://www.ign.com/articles/2019/09/18/check-out-baldurs-gate-1-2-and-icewind-dale-enhanced-edition-gameplay-on-ps4`

Blocking this script will prevent the ads from showing, and stop the site from breaking.

Ubo fixes this in their ublock origin extra extension.

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -283,7 +283,7 @@
 ! Anti-adblock: xiaomitoday.com
 @@||xiaomitoday.com/wp-content/themes/jannah/assets/js/advertisement.js$script,domain=xiaomitoday.com
 ! Anti-adblock ign.com
-||bwtjhnrr.g02.ign.com^$script,domain=ign.com
+||b8n5kh.g02.ign.com^$script,xmlhttprequest,domain=ign.com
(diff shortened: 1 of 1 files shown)


snyderp:
This seems likely to change often.  Can we generalize more broadly?  something like `.*\..*\.ign.com/[a-z0-9]{36}$script`?

ryanbr:
If we block the other scripts, it'll render the site blocken. (seems to affect the css). I would've also prefer a more generic fix, but they're doing some funky script dependency if we block one-too many scripts it'll cause rendering issues.

Will modify the patch to match the similar filter used in the Anti-CV list. (similar idea)

`https://github.com/abp-filters/abp-filters-anti-cv/blob/master/english.txt#L239`

PR #219, opened 2019-09-30
Simplify the reddit.com ad-tracking filter

ryanbr:
We probably don't need regex on this filter. This block the following script;

`https://www.redditstatic.com/desktop2x/js/ads.js`


snyderp:
SGTM 

Also, can you squish commits for PRs going forward?

PR #233, opened 2019-10-07
Anti-adblock on metacritic.com

ryanbr:
As seen on `https://www.metacritic.com/game/playstation-4/fifa-20/trailers/13567081`

(Testing on a US Vpn).

This specific url is launching other instart ad scripts, 
`https://c-9pruhskhx78v49x24vwdwlfx2ephwdfulwlfx2efrp.g00.metacritic.com/g00/3_c-9zzz.phwdfulwlf.frp_/c-9PRUHSKHXV49x24kwwsvx3ax2fx2fvwdwlf.phwdfulwlf.frpx2fvlwhqrwlfhx2f758x2fphwdfulwlfx2fvhwwlqjv.mvx3fl43f.pdunx3dvfulsw_$/$/$/$/$?i10c.ua=1&i10c.dv=21`

The block can't be generic, since it'll break the rendering of the site. Similar to the ign.com filter, https://github.com/brave/adblock-lists/pull/204

Submitted the same requested on anti-cv, https://github.com/abp-filters/abp-filters-anti-cv/pull/279

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -281,7 +281,9 @@
 @@||thetimes.co.uk/d/js/ads-$script,domain=thetimes.co.uk
 ! Anti-adblock: xiaomitoday.com
 @@||xiaomitoday.com/wp-content/themes/jannah/assets/js/advertisement.js$script,domain=xiaomitoday.com
-! Anti-adblock ign.com
+! Anti-adblock metacritic.com  (instart ads)
+||c-9pruhskhx78v49x24vwdwlfx2ephwdfulwlfx2efrp.g00.metacritic.com^$script,domain=metacritic.com
+! Anti-adblock ign.com  (instart ads)
 ||b8n5kh.g02.ign.com^$script,xmlhttprequest,domain=ign.com
 ! fix first-party items on maxmind.com (https://community.brave.com/t/stop-blocking-the-website-for-maxmind-com/68326)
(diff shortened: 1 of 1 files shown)


snyderp:
@ryanbr do you know if this URL is fixed / constant?  I imagine its DGA'ed, in which case we'd need some more general pattern, no?

ryanbr:
Its a static url, but it's likely it will change at same point. applicable to any instart-ads urls, they're static until they change it to get around any blocks.

ryanbr:
Just re-checked, the domains haven't changed. While they look random, they're constant at the moment. But as with any Anti-adblock,.. it'll change.

The "Ubo- Origin Extra" extension has a script "noop.txt" which is used on this site, and the following filters;

`/ad?i10c.getabdurl=*&ad_channel=$first-party,media`
`/g00/*$first-party,other`

But without the noop.txt snippet, the site will break with those filters alone.

ryanbr:
Okay we can wait. Though the alternative would be to push this out and then revert it once uBO resources fixes it.

snyderp:
I'm fine with the push-then-revert strategy if you can take charge of making sure the revert happens :)

If thats the case, just give me a thumbs up and I'll approve / merge.

ryanbr:
Site has since changed, no longer applicable. Closing.

PR #247, opened 2019-10-17
Anti-adblock on few more idg sites

ryanbr:
From; `https://www.techconnect.com/article/3441800/grab-a-kitchenaid-cold-brew-maker-for-80-on-amazon-today.html`
From: `https://www.pcworld.com/article/3432512/nvidia-gamescom-game-ready-driver-improves-performance-latency-and-sharpness.html`
From; `https://www.techadvisor.fr/long-format/gadgets/bon-plan-audio-3695492/`

**Scripts:**

`https://www.techconnect.com/www/js/ads/ads.js`
`https://www.techadvisor.fr/scripts/ads.js`
`https://www.pcworld.com/www/js/ads/ads.js`

**Source:**
`var canRunAds=true;`

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -328,6 +328,9 @@
 @@||phandroid.com/ads.js$script,domain=phandroid.com
 ! Adblock-Tracking: idg sites 
 @@||networkworld.com/www/js/ads/ads.js$script,domain=networkworld.com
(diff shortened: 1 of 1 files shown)


snyderp:
just noting uBO, since it would be good, going forward, to, instead of excepting things, replacing them with hard coded `var canRunAds=true;` and similar, to prevent them from changing under us

PR #250, opened 2019-10-18
Prevent Anti-adblock message on all vox sites

ryanbr:
Example, visiting `https://www.theverge.com/` would check for 2 scripts (concert.io related). If either item is blocked it would show the Anti-adblock message at the top of the page. Its also applicable on all vox sites.

`https://publicwww.com/websites/%22vox-cdn.com%22/`

More samples; `https://www.polygon.com/` `https://www.sbnation.com/` And all the sbnation blog (300+  sites); `https://www.sbnation.com/blogs`

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -201,8 +201,9 @@
 ||ttauri.space.com^$script,domain=space.com
 ||ttauri.tomsguide.com^$script,domain=tomsguide.com
 ||pcgamer-gb.pcgamer.com^$domain=pcgamer.com
-! Anti-adblock: theverge.com (vox sites)
-@@||vox-cdn.com/packs/concert_ads-$script,domain=theverge.com|eater.com|polygon.com|vox.com|sbnation.com|curbed.com
+! Anti-adblock: concert.io (vox sites)
+@@||vox-cdn.com/packs/concert_ads-$script,third-party
+@@||concert.io^*/concert_ads.js$script,third-party
 ! Adblock-Tracking: tweakers.net
 ||tweakimg.net/x/scripts/min/banners.js$script,domain=tweakers.net
 ! Adblock-Tracking: animeflv.net/animeflv.com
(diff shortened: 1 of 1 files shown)


snyderp:
@ryanbr but are a couple of global exceptions here, including to `@@||concert.io^*/concert_ads.js$script,third-party`, which worries me.  Can we figure out some other solution here, or limit it to some domains?

ryanbr:
I was considering a couple of fixes, remove concert.io outright or this, just too many vox (350 domains) to $domain= `https://publicwww.com/websites/%22cdn.concert.io%22/`

This is the url being check;

`https://cdn.concert.io/lib/concert-ads/v2-latest/concert_ads.js`

Okay, reconsidered. I've removed concert.io.

https://github.com/easylist/easylist/commit/81d7a939bdcdddd635617358102b99cedc2249e6

ryanbr:
I will gather the concert.io checks will be reversed. but that might be easier to deal with.

snyderp:
I think in this place, it'd be better to just block concert.io and deal with the anti-adblock message, since it seems very likely that concert.io is tracking people.  WDYT?

https://concert.io/

ryanbr:
Okay, how about I add the following block to EP;

`https://cdn.concert.io/lib/concert-concierge.2.4.0.min.js`

`||concert.io/lib/$script,third-party`

Avoiding the need to whitelist work arounds for `https://cdn.concert.io/lib/concert-ads/v2-latest/concert_ads.js` Which seems to be only used for Adblock checks.

snyderp:
That sounds promising!  Can you update the PR with that, so I can review the final proposal?  But sgtm if im following right

ryanbr:
https://github.com/easylist/easylist/commit/ae19dc8d8d6158c2a0d9a60e0433682ca1f8bf78

This will block the non-adblock checking script.  The current pull request should be fine.

snyderp:
@ryanbr i'm still nervous about adding the global exception for what looks like it could cover a lot of things, and not just one particular anti-ad-block script.

`https://cdn.concert.io/lib/concert-ads/v2-latest/concert_ads.js` seems to do non-trivial tracking too, including some cookie resync (if im reading it right).  What if we leave the block in place (so leave the anti-adblock message unaddressed for now) and re-tackle once the uBO stuff is in place?

ryanbr:
We just pick the same popular domains like last time, I guess. Seems ubo stuff didnt cover all the vox domains either.

PR #251, opened 2019-10-19
Update Anti-ad tracking on healthline.com

ryanbr:
Just a minor adjustment. Script folder was changed. Visiting `https://www.healthline.com/` will show the updated check.

`https://assets.healthline.com/content/advertising.js`

Blank file check, no source.


snyderp:
Can we limit this to certain domains too? 

In general, by default, all exception rules should be limited to a single domain.  If we want to unblock something broader, we should talk through and or measure

ryanbr:
Yeah I just edited an exisitng line, missed the $domain switch.

On related issue, https://github.com/brave/adblock-lists/pull/252  A cleanup of these filters missing switches.

PR #271, opened 2019-11-02
Add various Anti-adblock Instart ads websites

ryanbr:
This is combined fix for a few related Instart domains; I was hoping the previous patches would be more static, but alas no.

2 new domains;

`https://www.pockettactics.com/`

`https://www.tribunnews.com/`

And updates on `ign.com` and `webmd.com`. Gamespot is still fine.


snyderp:
Glad to update these, though do you have a sense of how quickly those has-or-DGA-style subdomains change?  If its regularly, would it make sense to just automate things?  Like create a separate list for filter rules targeting URLs that regularly change, and just set up some bots to automatically check and re-gen?

PR #273, opened 2019-11-06
Fix anti-adblock on reuters.com

ryanbr:
`reuters.com` added a new check on the site. This will stop the dianomi ads from being loaded.

As seen on  `https://www.reuters.com/article/us-usa-election/democrats-claim-victory-over-trump-backed-kentucky-governor-seize-virginia-legislature-idUSKBN1XF1EU`

`https://www.reuters.com/ads.js`

Source:

`var e=document.createElement('div');`
`e.id='ApoHvDqryePk';`
`e.style.display='none';`
`document.body.appendChild(e);`


snyderp:
@ryanbr looks like this has gone stale or something.  Can you fix the conflicts and lets update from there?

ryanbr:
Done :) @snyderp

PR #277, opened 2019-11-09
Add more admiral servers

ryanbr:
Just a sync Admiral server list with Easyprivacy (ios) Just a followup from the previous pull req

https://github.com/brave/adblock-lists/pull/276

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -447,6 +447,16 @@
 ! Admiral Tracking (for ios)
 ||acridtwist.com^$third-party
 ||caringzinc.com^$third-party
+||ringplayground.com^$third-party
+||jamexistence.com^$third-party
+||roofrelation.com^$third-party
+||gulliblecamp.com^$third-party
+||puffypull.com^$third-party
+||stupendoussleet.com^$third-party
+||burnbubble.com^$third-party
+||butterbulb.com^$third-party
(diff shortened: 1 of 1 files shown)


snyderp:
@ryanbr maybe it'd be best to make a separate, iOS suplement list?  We can bundle that with the next iOS release, and that way other platforms don't get duplicate rules?  WDYT?

ryanbr:
When we properly sync with Easyprivacy in slimlist, these rules could be removed.

snyderp:
not necessarily, since its possible slimlist would not come across these domains during its random crawl.  So my thinking is that it'd be useful to both have a list of things that were always ship to iOS (and only iOS) regardless of SlimList, for this reason, but also maybe for web compat reasons too

ryanbr:
Yeah, wouldn't be hard to move these rules into its own "admiral list" for slimlist.

snyderp:
Sure, can you create that separate list?  And name it something more general, so that we can use it for other similar needs without the same naming problem we got into with `brave-unbreak`?  So something like `brave-ios`?

@jhreis whats the best way of getting this iOS filter list supplement integrated?

ryanbr:
https://github.com/brave/adblock-lists/issues/279

ryanbr:
PR was superseded by https://github.com/brave/adblock-lists/pull/326

PR #281, opened 2019-11-11
Fix video playback on https://www.ctv.ca/The-Rookie

ryanbr:
Blocking the following 9c9media.ca is breaking the site (from the disconnect list); as seen on `https://www.ctv.ca/The-Rookie`

`https://license.9c9media.ca/widevine`
`https://auth.9c9media.ca/auth/main.js`

Maybe affecting other Canadian sites, without the whitelists video playback is broken. I didn't see any tracking from this 9c9media.ca domain. Seems to be used for licensing ?


3052:
note this breaks videos now, as widevine is required on that site now

PR #292, opened 2019-11-22
Anti-adblock tracking on https://abc.com/

ryanbr:
Anti-adblock tracking from; `https://abc.com/shows/jimmy-kimmel-live/video/featured/vdka14524148`

**Script:**
`https://cdn1.edgedatg.com/aws/apps/datg/web-player-unity/1.1.8.15/js/ads.min.js`

**Source:**
`var e=document.createElement("div");e.id="LkFtUuy",e.style.cssText="display:block; width: 0px !important; height: 0px !important",e.className="ads-container",document.body.appendChild(e);`

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -431,6 +431,8 @@
 ! Anti-adblock: washingtonpost.com
 @@||d2ty8gaf6rmowa.cloudfront.net/ad/$domain=washingtonpost.com
 @@||pubads.g.doubleclick.net^$xmlhttprequest,domain=washingtonpost.com
+! Anti-adblock tracking: abc.com
+@@||edgedatg.com^*/ads.min.js$script,domain=abc.com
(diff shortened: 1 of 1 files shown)


snyderp:
@ryanbr just a reminder, please mark these for post ubo parity stuff, so we can lock them down

ryanbr:
@snyderp ubo don't seem to cover adblock tracking (like this). I did check their privacy.txt, filters.txt, nothing in the list for abc.com/edgedatg.com. Seems to more a issue when it actually notifys or limits users on the site. In this case, its seems to be just tracking users (which could change).

snyderp:
I dont mean that uBO has a solution to this, i mean that once uBO resource replacement is live on all platforms, we should use that capability to lockdown the scripts we're excepting (so that they don't get swapped out from underneath us)

PR #297, opened 2019-11-28
Include anti-notifications list

ryanbr:
Imported list from:

`https://github.com/easylist/easylist/blob/5d767f3cb6bc11d0f7ae1be4edfa0305141ab4af/fanboy-addon/fanboy_annoyance_specific_block.txt#L59`

`https://github.com/easylist/easylist/blob/5d767f3cb6bc11d0f7ae1be4edfa0305141ab4af/fanboy-addon/fanboy_annoyance_thirdparty.txt#L409`

We could include the generic push notifications scripts also if needed.


snyderp:
@ryanbr can you add this new list to https://github.com/brave/adblock-rust/blob/master/src/filter_lists/default.rs as a PR against https://github.com/brave/adblock-rust?

Most of those fields should be self explanatory I believe.  For the UUID, you can just generate a new one

PR #300, opened 2019-12-06
Add 3rdparty script govdelivery.com

ryanbr:
Visiting `https://www.braintree.gov.uk/`

Will bring up a overlay on the site, from `govdelivery.com` Can't really add this to Easyprivacy, but its worth adding here.

`https://content.govdelivery.com/overlay/js/4025.js`


ryanbr:
Also seen on `https://www.breckland.gov.uk/`

PR #301, opened 2019-12-07
Fix ads on https://rutracker.net/forum/index.php

ryanbr:
Possibly related to this bug; https://github.com/brave/brave-browser/issues/5440 a user reported ads showing up on the rutracker site (https://twitter.com/lukemulks/status/1203421141978644480)

Should limit some of the ads being loaded.

Same filters from https://easylist-downloads.adblockplus.org/advblock.txt
`||rutrk.org/`
`||betsonsport.ru^$third-party`


snyderp:
I don't see any relationship to https://github.com/brave/brave-browser/issues/5440 (which is about unintended consequences of exception rules from sources other than the main shared DAT) but this is approved either way.

BalbuBlock:
@snyderp  It may be necessary to review the rules of exceptions if they block the blocking of ads on this kind of site, right? Thanks for the merged.

PR #304, opened 2019-12-16
Fix uBo list addition block on tags.tiqcdn.com

ryanbr:
Have previously experimented blocking this script before, and it'll cause many false positives. 

Initial False positive:
`https://community.brave.com/t/hyperlink-does-not-work/98608` 

Only recently committed 2 days ago. Have requested a revert: https://github.com/uBlockOrigin/uAssets/commit/dc12acd80d1d64be596272c66ceb0b7925828347 

But not sure how long this will take for uBO authors, or if they'll revert this.


ryanbr:
Closed, was reverted.

PR #314, opened 2020-01-13
fixes brave/brave-browser/5406 (WIP)

ryanbr:
This a quick fix on an outstanding issue on https://github.com/brave/brave-browser/issues/5406 for `skepticalscience.com`

Blocking this script will prevent the browser from a lock up. This solution is a temp fix, and could be re-tested/removed in future.

Was re-reported here: https://community.brave.com/t/why-does-this-website-lockup-the-browser/102074/2

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -34,6 +34,8 @@
 ! theatlantic.com anti-blocker filters
 ||theatlantic.blueconic.net$domain=theatlantic.com
 ||theatlantic.com/please-support-us^
+! Fix browser lockup on skepticalscience.com (https://github.com/brave/brave-browser/issues/5406)
+||skepticalscience.net/widgets/heat_widget/js/heat_content.js$script,domain=skepticalscience.com
 ! adops.com unusable without this
 @@||adops.com^$~third-party
 @@||www.scrumpoker.online^$~third-party


snyderp:
@ryanbr I appreciate the goal here, but I don't think we should merge this if the relevant script is tracking related, w/o at least a better solution planned out.  Is this something we could solve with a uBO style replacement?

In ohther words, I'm happy to merge this, but only if @ryanbr you can take it on as an action item to make sure we get a better solution worked out here in the short term (even if its not on you to implement the better solution).  WDYT?

ryanbr:
I'll investigate a better option, atleast we know which script is causing it now.

diracdeltas:
alternatively we could just allow fingerprinting from skepticalscience.net on skepticalscience.com in https://github.com/brave/brave-core/blob/master/common/shield_exceptions.cc#L41. that might fix it given that the site works fine without fp turned on.

ryanbr:
The ubo method for reference: `skepticalscience.net##+js(aopr, hcdrawticks) `

snyderp:
@ryanbr @diracdeltas sounds like we have a way forward then?  Keep the just-added exception in place until cosmetic filtering lands and hits stable, and then pull it out?

If thats right, @ryanbr can you keep that on your todo list, to pull the just-added exception once cosmetic filtering hits stable?

PR #320, opened 2020-01-18
Fix embedded reddit.com videos

ryanbr:
Visiting: `https://www.dexerto.com/overwatch/insane-overwatch-exploit-makes-widowmaker-bulletproof-on-gibraltar-1314840` Preventing an embedded reddit video clip from showing.

The following items are blocked incorrectly:

`https://www.reddit.com/comments/eqelnw.json?limit=1&app=embed`
`https://old.reddit.com/mediaembed/eqelnw`

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -62,7 +62,7 @@
 ! https://github.com/brave/adblock-lists/issues/39
 @@||alb.reddit.com^
 ! Allow sites to use reddit api
-@@||reddit.com^*/.json?$script,xmlhttprequest,third-party
+@@||reddit.com^*.json?$script,xmlhttprequest,third-party
 @@||reddit.com^*/access_token$xmlhttprequest,third-party
 @@||oauth.reddit.com^$xmlhttprequest
 @@||reddit.com/mediaembed/$subdocument,third-party
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -65,6 +65,8 @@
 @@||reddit.com^*/.json?$script,third-party
 @@||reddit.com^*/access_token$xmlhttprequest,third-party
 @@||oauth.reddit.com^$xmlhttprequest
+@@||reddit.com/comments/$xmlhttprequest,domain=ampproject.net
+@@||reddit.com/mediaembed/$subdocument,domain=ampproject.net
 ! reddit share buttons
 @@||reddit.com/static/button/$subdocument,third-party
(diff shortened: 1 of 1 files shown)


snyderp:
@ryanbr would this make sense to just globally allow?  If the contents of that AJAX request doesn't contain any identifying material, maybe that'd be better.  WDYT?

ryanbr:
was being cautious, but yeah seems fine to make a little less domain specific. And combined the json request.

snyderp:
Is this something that could be upstreamed?

ryanbr:
Unless the disconnect list guys remove the Reddit block, probably not. Should we prod them?

snyderp:
@ryanbr I think its fine to ask them, but only if you're confident you know which requests to which reddit related domains are tracking related, and which aren't.  I expect disconnect folks are right here, i thought the issue was with EasyList/EP. 

We're in a different position than most disconnect consumers b/c we never send cookies. 

Before I merge in the above change though, can you take a look at the AJAX calls this would un-block and see if there looks like any tracking-related material in the contents of the AJAX bodies?  If anything looks tracking related, then we might need to switch to a polyfill.  But if you have a good degree (doesn't need to be 100% certainty) of confidence that its not state that describes the user, just state that describes the page and thing being requested, then I'll merge this in now

ryanbr:
The json seems pretty sane: https://pastebin.com/eJ1ZVZXd  (From the blocked request `https://www.reddit.com/comments/eqelnw.json?limit=1&app=embed`) Most/if not every json variabes are just reddit variables.

The only questionable variable
`        "whitelist_status" : "all_ads",`

Disconnect list bascially blocking any reddit, out side of reddit.com `https://github.com/brave/adblock-lists/blob/master/brave-disconnect.txt#L1597`

`||reddit.com^$third-party`

We could $badfilter as an option, From experience, all the tracking on reddit is done on reddit.com domain (first-party tracking).

`easyprivacy_specific.txt:||alb.reddit.com^`
`easyprivacy_specific.txt:||events.reddit.com^`
`easyprivacy_specific.txt:||events.redditmedia.com^`
`easyprivacy_specific.txt:||gateway.reddit.com|`
`easyprivacy_specific.txt:||pixel.reddit.com^`

Some/most of these sub domains maybe dead now.

snyderp:
@ryanbr if it looks to not be identifying the user, and you feel confident it won't enable tracking, then that all sounds terrific :) Thanks for looking into this!

PR #322, opened 2020-01-24
Fix Non-clickable links on https://www.elconfidencial.com/

ryanbr:
This affected Brave specifically than ubo by itself. White listing EventTracker.js by ubo is causing links to be non-clickable

Visit: `https://www.elconfidencial.com/` 
With shields enabled, no links will be clickable.

Letting Easyprivacy blocking `/EventTracker.js` should fix this site.

I was going to land a PR removal on uBO but I couldn't reproduce the same issue with uBO lists enabled.

Closing this. causing another issue with comments.


ryanbr:
Okay, a bit more digging. Its a cookie consent issue. Again, while this patch fixes the clickable issue, still breaks comments.

PR #345, opened 2020-03-02
Fix https://suumo.jp/ on IOS

ryanbr:
Filter is taken from https://github.com/easylist/easylist/commit/04d3f7179ce7

Url we're whitelisting: `https://asset01.suumo.jp/sp/js/beacon.js?id=2020022728cf47abcd`

Visiting: `https://suumo.jp/sp/chintai/tokyo/ek_004026540/cond/?sjoken%5B0%5D=059` wouldn't let the user go the next page at the bottom.

Affecting IOS as reported by @chkk525

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -148,6 +148,8 @@ stats.brave.com#@#adsContent
 @@||auth.9c9media.ca^$script,domain=ctv.ca
 ! 2mdn video playback script
 @@||2mdn.net/instream/html5/ima3.js$script,domain=zdnet.com|techrepublic.com
+! suumo.jp (ios)
+@@||suumo.jp/sp/js/beacon.js$script,domain=suumo.jp
 ! ups.com fix (ios)
 @@||usps.com/go/scripts/tracking.js$script,domain=usps.com
 ! Adblock-Tracking: theintercept.com


pes10k:
@ryanbr would it make sense to add iOS only list?  Do you have a sense of how many rules are only targeting iOS?

ryanbr:
Not sure if it would be needed (since only cover small / very specific circumstances). Maybe Anti-adblock (Which Brave Android also benefits), and `brave://adblock` requests https://github.com/brave/adblock-lists/pull/266 https://github.com/brave/adblock-lists/pull/85

Maybe we can create brave-mobile list, to cover both ios/android. and move most/if not all Anti-adblock into this list?

From https://github.com/brave/adblock-lists/issues?utf8=%E2%9C%93&q=label%3A%22ios+specific%22  

**Just to detail why we have the following filters:**

Some Anti-adblock filters would target ios/Android (Lack of snippet support): https://github.com/brave/adblock-lists/pull/85  

Since IOS dosn't have `brave://adblock`: https://github.com/brave/adblock-lists/pull/266

If we're pulling/converting the changes from Easyprivacy, we could remove: 
https://github.com/brave/adblock-lists/pull/302
https://github.com/brave/adblock-lists/pull/326
https://github.com/brave/adblock-lists/pull/276
https://github.com/brave/adblock-lists/pull/277
https://github.com/brave/adblock-lists/pull/345 (this commit)

pes10k:
That all sounds good to me.  I'll merge this in now, but @ryanbr can you keep that on your radar, and when the number of ios / android / mobile specific rules get to say ~50, lets do it.  In the meantime, just noting which rules are targeting mobile should be good

PR #349, opened 2020-03-03
Remove redirect badfilter, as we support scriptlet redirects now

antonok-edm:
@ryanbr this should fix businessinsider's popup modal


ryanbr:
Okay, looks fine.

PR #352, opened 2020-03-21
Hide generic elements on forbes.com

ryanbr:
Since we don't have `$generichide` currently, `forbes.com` is checking on these cosmetic elements. 

This will prevent the Anti-adblock message from showing until $generichide is supported.


ryanbr:
**uBO currently using:** 

```
@@||forbes.com^$generichide
forbes.com##main:style(padding-top:10px!important)
forbes.com##[class^="fbs-ad"]
```
The first 2 filters will prevent the warning from showing.

BrendanEich:
Thanks!

PR #353, opened 2020-03-24
Fix https://ca.yahoo.com/

ryanbr:
Will prevent inline ad-videos on `https://ca.yahoo.com/` from loading (Using a .ca VPN)

Ad elements picked up.

`https://s.yimg.com/uu/api/res/1.2/V7gU.rpV5gRXgNst41gp4g--~B/Zmk9c3RyaW07aD0zOTI7cHlvZmY9MDt3PTc1MDtzbT0xO2FwcGlkPXl0YWNoeW9u/https://s.yimg.com/av/ads/1584853238138-4462.jpg.cf.webp`

`https://s.yimg.com/uu/api/res/1.2/GrsUjITQxlghAL0axNtIBg--~B/Zmk9c3RyaW07dz0xNjA7aD0xNjA7YXBwaWQ9eXRhY2h5b24-/https://s.yimg.com/av/ads/1539205411300-3868.jpg`

Also committed fixes in EL: https://github.com/easylist/easylist/commit/c2479372470c2ea4e6fc3b29d4ffac82b6403b5a

If we sync with Easylist in Ios, it should pick up my changes.

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -147,6 +147,8 @@ stats.brave.com#@#adsContent
 @@||auth.9c9media.ca^$script,domain=ctv.ca|crave.ca
 ! 2mdn video playback script
 @@||2mdn.net/instream/html5/ima3.js$script,domain=zdnet.com|techrepublic.com
+! ca.yahoo.com (ios)
+/av/ads/*
 ! suumo.jp (ios)
 @@||suumo.jp/sp/js/beacon.js$script,domain=suumo.jp
 ! ups.com fix (ios)
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -148,7 +148,7 @@ stats.brave.com#@#adsContent
 ! 2mdn video playback script
 @@||2mdn.net/instream/html5/ima3.js$script,domain=zdnet.com|techrepublic.com
 ! ca.yahoo.com (ios)
-/av/ads/*
+/av/ads/*$domain=yahoo.com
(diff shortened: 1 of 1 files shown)


pes10k:
This is pretty broad.  Is there a way to narrow to prevent false positives?

ryanbr:
I just copied/pasta from the commit in EL, but yeah I can understand. Made more specific.

pes10k:
one last nit; is the `*` needed?

ryanbr:
Correct, the wildcard at the end is needed (well, commonly used in Easylist)

pes10k:
hmm, what URL would `/av/ads/*` match that `/av/ads/` wouldn't match though?

nitting about this bc (I believe) `*` triggers the rule to be applied as if its a regular expression, which is slower

ryanbr:
In Easylist, if a filter starts with "/" and then also ends with "/", it will need end with a wild-card. Otherwise it's considered a "slow filter".

To test the rule, visit; https://abpvn.com/ruleChecker/redundantRuleChecker.html   

Under **Warnings**
`The following error, warning or optimalization was encountered while checking the rules:`
`/av/ads/ : Unnecessary regular expression. Use 'av/ads' instead, or use '/av/ads/*' if it isn't a regex`

I could change it if need be, just keeping the same formatting requirements as Easylist

pes10k:
hmm, odd!  guessing its going to be engine dependent then.  I think (but am not sure) that adblock-rust would do this different, but, not at all important at the moment.  Thanks!

PR #372, opened 2020-04-21
Cleanup and remove all youtube related filters from ABP-Japanese

ryanbr:
Basically I've removed combined every filter that could still be potentially causing issues (still being reported via ABP Japanese)


ryanbr:
A bit excessive for sure, but this has been ongoing for too long. ABP Japanese issues have kept going.

Easylist covers all the generics, without ABP Japanese creating bad generics.

PR #373, opened 2020-04-23
Revert https://github.com/brave/adblock-lists/commit/69267f8a4905db64d725da1abe2d09ed3591121e

ryanbr:
revert previous commit.


ryanbr:
The filters should've worked (but something broke), this is a revert to see whats going on

PR #378, opened 2020-05-03
Fix facebook game Ludo King

ryanbr:
From `https://www.facebook.com/instantgames/play/2264592520437327/?context_type=SOLO&source=fb_gg_url`

We're incorrectly blocking `||connect.facebook.net/en_US/fbinstant.6.2.js$script,domain=apps-2264592520437327.apps.fbsbx.com`


pes10k:
@ryanbr Please do not make modifications or restrictions on rules with `tag=fb-embeds` or similar exceptions.  These rules are what power / enable the social media blocking settings on the back end.

We should break anything that requires FB when `tag=fb-embeds` rules are enabled

PR #383, opened 2020-05-10
Fix/unblock liveperson.com 

ryanbr:
`||liveperson.com^$third-party,domain=~liveperson.net`

Due to this block, it'll prevent web commerce sites from using web chat communication between them and customers.

`https://www.lowes.com/` and `https://www.darphin.com/` uses liveperson.com/.net (and many https://publicwww.com/websites/%22liveperson.com%22/) 

Interesting enough Disconnect list does not block `liveperson.net`, which is doing the same thing.

We have two blocks on these domains in EL/EP:
`easylist/easylist_thirdparty.txt:||liveperson.com/affiliates/`
`easyprivacy/easyprivacy_specific.txt:||liveperson.net/hc/*/?visitor=`

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -530,6 +530,8 @@ hardwareluxx.de,formel1.de,reuters.com,golem.de,finanzen.net,autobild.de,gamesta
 @@||plus.google.com/js/$script,domain=shapeways.com
 ! Anti-adblock tracking: abc.com
 @@||edgedatg.com^*/ads.min.js$script,domain=abc.com
+! Fix liveperson.com (Disconnect block) breaks website chat
+||liveperson.com^$badfilter,third-party,domain=~liveperson.net
 ! Fix Costco (Disconnect block on channeladvisor.com)
 @@||channeladvisor.com/ImageDelivery/$image,third-party
 ! Fix yac.chat (Disconnect block on viral-loops.com)


ryanbr:
Have requested removal https://github.com/disconnectme/disconnect-tracking-protection/issues/176

pes10k:
@ryanbr is there anything we could do to block any tracking aspects of the service, while still allowing the user serving parts?

ryanbr:
`https://static-assets.dev.fs.liveperson.com/lowes/gk.js` seems okay, this is the script being impacted.

The script `https://lptag.liveperson.net/tag/tag.js?site=22554410` has a UA check, however the .net domain isn't blocked. If we block this script, it'll prevent the other scripts from loading.

pes10k:
Sorry, im not following your reply.  Is there a narrower rule we can use?

ryanbr:
Based on the disconnect block on `liveperson.com`, I see no privacy issues with `https://static-assets.dev.fs.liveperson.com/lowes/gk.js`. 

Given the popularity of `liveperson.com` a whitelist shouldn't be used. https://publicwww.com/websites/%22liveperson.com%22/  

I got another report today, that `americanexpress.com` chat is also having issues, they're also using liveperson.

pes10k:
okie okdie, if you're confident this is all non-privacy harming, this all seems good to me.  THanks!

PR #386, opened 2020-05-23
Allow apps.fbsbx.com to use facebook domains

ryanbr:
Visit: https://www.facebook.com/instantgames/play/ click on `Play`, expecting games to load. 
Will cause endless loading.

Allowing:
`@@||connect.facebook.com^$script,domain=apps.fbsbx.com`
`@@||connect.facebook.net^$script,domain=apps.fbsbx.com`

This should be safe, its only contained within facebook's own domain `apps.fbsbx.com`, which is not used externally outside of facebook.com.

I prefer not to whitelist `connect.facebook.com` outright, this is limited to facebook's own domain only. No trackers are whitelisted and are still blocked. Hopefully bring down false positives reported in our stats.

fb-requests
fb-requests2


ryanbr:
Okay, seems to be false positive on my end. not actually needed.

PR #400, opened 2020-06-09
Sync With NoCoin List and nocoin-ublock

ryanbr:
1. Removed redundant custom filter `||load.jsecoin.com$third-party`  (redundant because of `||jsecoin.com^$third-party`)
2. Kept: `https://github.com/brave/adblock-lists/pull/32` `||greenindex.dynamic-dns.net/jqueryeasyui.js^$third-party`
3. Also properly import uBO no-coin snippets from https://github.com/hoshsadiq/adblock-nocoin-list/blob/master/nocoin-ublock.txt

### DIFF
--- a/coin-miners.txt
+++ b/coin-miners.txt
@@ -10,22 +10,21 @@
 ! Description: This filter disables browser based miners such as coin-hive. Many rules borrowed from https://github.com/hoshsadiq/adblock-nocoin-list
 !-------------------------------------------------------------------------------
 
-! Custom rules
-||load.jsecoin.com$third-party
-
 ! Third Party blocking ---------------------------------------------------------
 ||cnhv.co^$third-party
 ||coin-hive.com^$third-party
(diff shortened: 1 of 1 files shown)


ryanbr:
Given the size of the merge, If you're okay with this @pes10k  :)

pes10k:
@ryanbr can you confirm that the new no-coin rules aren't redundant with https://github.com/brave/adblock-lists/blob/master/coin-miners.txt ?

ryanbr:
This will actually update (all) the filters in https://github.com/brave/adblock-lists/blob/master/coin-miners.txt    Just checked via a redundancy on https://abpvn.com/ruleChecker/redundantRuleChecker.html  no redundant filters, some dubes in Easyprivacy at most. @pes10k

pes10k:
Ah, then terrific, lets make it so :)

PR #423, opened 2020-07-23
Include Brave social disconnect

ryanbr:
Having a few versions of a cut-down version of the previous `brave-disconnect.txt`. This is the better way to go:

1. We're not reliant on unknown reasons on why a filter is in the list
2. Avoiding over blockage of non-tracking domains.
3. We're covering our needs only what is part of `brave://settings/socialBlocking`
4. Many of the domains aren't used for tracking or we already block them via existing generic filters in https://github.com/easylist/easylist/blob/master/easyprivacy/easyprivacy_general.txt anything missing will added to Easyprivacy as we normally address this.

This list could be trimmed further (removing some of the google domains already in EL/EP). But this is a good base to start with.

### DIFF
--- a/brave-lists/brave-social.txt
+++ b/brave-lists/brave-social.txt
@@ -1,0 +1,15 @@
+! List used by Brave for preventing social elements from loading
+!
+! Facebook
+||connect.facebook.net^$third-party,domain=~atlassolutions.com|~facebook.com|~facebook.de|~facebook.fr|~facebook.net|~fb.com|~fb.me|~fbcdn.net|~fbsbx.com|~friendfeed.com|~instagram.com|~internalfb.com|~messenger.com|~oculus.com|~whatsapp.com|~workplace.com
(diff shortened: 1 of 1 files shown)


ryanbr:
- Removed `||gmail.com^$third-party`  and `||googlemail.com^$third-party` (Not used outside of gmail, also domain change to mail.google.com)
- Removed `||google-analytics.com^$third-party`, In Easyprivacy.
- Removed `||googlesyndication.com^$third-party`, In Easyprivacy
- Removed `||googletagservices.com^$third-party` In Easyprivacy

ryanbr:
Removed google filters, which isn't doing much since we're blocking google-analytics, and other google tracking domains

ryanbr:
Cleaned up facebook domain to only public facing 3rd/party ||connect.facebook.*. will improve compatibility with facebook games

ryanbr:
Removed `vk.com` related domain, while a social server, not a tracker and no way to get around the in brave social settings for this domain. We're already blocking the tracking bits for `vk.com` already

PR #441, opened 2020-08-12
Remove redundant filters

ryanbr:
Filters for expressen.se and aftonbladet.se are no longer needed, confirm none of these filters are being used on either website.

Easylist, Easyprivacy + uBO has taken care of most, if not all of the issues.


ryanbr:
None of these filters are now used, checked each site, no hits. Site has changed, filters also changed.

ryanbr:
Added new twitch.tv filter , https://github.com/easylist/easylist/commit/eda63ebe110381f294d8bfbbb61803876fca0971

PR #450, opened 2020-08-18
Fix clickthrough for https://go.nordpass.io/aff_c?offer_id=000&aff_id=000

ryanbr:
Fix another clickthrough issue in ios. `https://go.nordpass.io/aff_c?offer_id=000&aff_id=000`

Caused by `"?offer_id=*&aff_id="`, this will let ios click on links without it causing issues. Refined from https://github.com/brave/adblock-lists/pull/449

Fixing `https://go.nordpass.io/aff_c?offer_id=000&aff_id=00000`
and `https://proton.go2cloud.org/aff_c?offer_id=00&aff_id=000`


ryanbr:
(on the related note, we removed the related filter in Easylist, not a privacy issue) Only used as a click through.

PR #463, opened 2020-09-15
Block requests to localhost and similar, when not on localhost and similar

### LINKED ISSUE #462: Block connections to localhost resources from non localhost pages
We should block websites connecting to local resources (localhost, 127.0.0.1, etc) that are no local themselves

pes10k:
fixes #462

### DIFF
--- a/brave-lists/brave-specific.txt
+++ b/brave-lists/brave-specific.txt
@@ -1,1 +1,4 @@
 ! Specific filters (Tracking or ads) for Brave
+||localhost^$third-party,domain=~127.0.0.1|~[::1]
+||127.0.0.1^$third-party,domain=~localhost|~[::1]
+||[::1]^$third-party,domain=~localhost|~127.0.0.1


llacb47:
add an anchor `||` ?

ryanbr:
@antonok-edm uBO doesn't like ipv6, can adblock-rust work with ::1?

llacb47:
change `~*.local` to `~local`. IDK if brave's adblock engine works differently but it works in uBo/abp

pes10k:
thanks @llacb47 , like i mentioned to @ryanbr, glad to have you both helping me brute force this (you can probably tell i dont write non-trivial filter list rules often ;)

ryanbr (approved):
If `::1` is fine then this is okay.

antonok-edm (changes requested):

ryanbr:
I would also consider changing from .local^$.. to ||local^$...

antonok-edm:
oh also, I don't think `~local` is working correctly - will continue to investigate

pes10k:
okie dokie, will hold off on more changes until you report back :)

pes10k:
okie as discussed, going to solve this by splitting into two PRs, this one w/o .local, and https://github.com/brave/adblock-lists/pull/465 which adds back in the .local addresses

antonok-edm (approved):

PR #465, opened 2020-09-16
add .local addresses in when detecting local resource requests

pes10k:
second half of fix for #462


antonok-edm (commented):
Blocked on https://github.com/brave/adblock-rust/issues/138 (which should make it into Brave release before we update this)

ryanbr:
@pes10k Quite old request, safe to close or should I update with a new PR?

pes10k:
I think this would be a good change, but best to check with @ShivanKaul if its still useful and necessary

antonok-edm:
@ShivanKaul says we don't need it since we have localhost permissions

PR #466, opened 2020-09-16
Revert "Block requests to localhost and similar, when not on localhost and similar"

pes10k:
Reverts brave/adblock-lists#463


pes10k:
Reverting until we have a solution for non-standard ports (currently adblock-rust will treat localhost diff from localhost:<port>, and we don't have a generic way of saying "localhost with all ports"

PR #468, opened 2020-09-17
block requests to local resources from non-local origins

pes10k:
re-doing https://github.com/brave/adblock-lists/pull/463 now that we know `localhost^` matches all ports on localhost


antonok-edm (approved):
nice

PR #473, opened 2020-09-24
Revert "Remove youtube js snippets"

ryanbr:


ryanbr:
Closed request, doesn't need to be reverted.

ryanbr:
Was going to revert for IOS. but won't be needed

PR #475, opened 2020-09-29
Address supleks.jp reinjection on mobile (fixes https://github.com/Ad…

Yuki2718:
…guardTeam/AdguardFilters/issues/64521)

Details:
https://github.com/AdguardTeam/AdguardFilters/issues/64521

Reproduced on Brave Android without AG:

ramendb

I think `##div[class*="_"][data-height]` is too broad for generic (likely to cause FP) so skipped to report to EL. I'm not sure if we should add cosmetic rules for some placeholder now or should wait for mobile support of cosmetic filtering.

Off-topic: if it was uAssets I would have added `!#if  env_mobile` but I guess it's not supported on Brave, right?


ryanbr:
@Yuki2718  JS Snippets isn't available in mobile for Brave just yet

Yuki2718:
Ah, okay then we don't have a solution until the day it supports at least cosmetic filtering.

PR #482, opened 2020-10-04
Fix 127.0.0.1 port scans on ebay-kleinanzeigen.de

ryanbr:
Fixes rolled out to Easyprivacy. https://github.com/easylist/easylist/commit/6216339d30c1d8162d20b166d8eecf7061ebcf1b

But this will also cover if ebay changes the script again. 

Was reported here: https://old.reddit.com/r/brave_browser/comments/j4tggj/this_ebay_site_freezes_brave_for_several_seconds/


pes10k:
@ryanbr is this still needed?  Does this prevent any port scanning that https://github.com/brave/adblock-lists/blob/master/brave-lists/brave-specific.txt misses? 

Asking bc if yes, then we should update the general portscan rules too

ryanbr:
@pes10k 

Yeah, it's still needed. This patch prevents this specific port scan script on ebay from initiating in the first place (which will help with performance).

The localhost blocks in https://github.com/brave/adblock-lists/blob/master/brave-lists/brave-specific.txt  are still applicable for ebay (and other sites), we don't need to change anything here.

PR #490, opened 2020-10-26
allow audited yandex code when on yandex sites

pes10k:
Allows Yandex to load some first party scripts, after auditing by brave.  This change does not affect storage policy or result in any 3p storage being enabled / allowed

re https://github.com/brave/brave-browser/issues/12327
after review in https://github.com/brave/brave-browser/issues/11591

cc @bsclifton

### DIFF
--- a/brave-lists/brave-specific.txt
+++ b/brave-lists/brave-specific.txt
@@ -2,3 +2,15 @@
 ||localhost^$third-party,domain=~127.0.0.1|~[::1]
 ||127.0.0.1^$third-party,domain=~localhost|~[::1]
 ||[::1]^$third-party,domain=~localhost|~127.0.0.1
+
+! Re-enable 1p scripts for Yandex after finding no privacy concerns for
+! Brave users.
+! Privacy Review: https://github.com/brave/brave-browser/issues/11591
+! Main Issue: https://github.com/brave/brave-browser/issues/12327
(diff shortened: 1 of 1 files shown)


pes10k:
The original version of this PR attempted to only enable this exception when yandex origins were the first party, but since there isn't a way to note TLDs in adblock rules, there wasn't a way to restrict this to yandex.<TLD> but not. yandex.tracker.com.  Can revisit if adblock syntax allows this in the future

ryanbr:
@pes10k 

Can we limit the filters too yandex.ru and yandex.com `$domain=yandex.com|yandex.ru` or all the yandex domains. `yandex.by|yandex.com|yandex.kz|yandex.ru|yandex.ua`

`https://mc.yandex.ru/metrika/watch.js`  seems to be doing fingerprinting (found on https://yandex.ru/search/?lr=113851&text=test)

pes10k:
@ryanbr sure, thats no problem at all.  My initial thinking here was that the type of fingerprinting done in https://mc.yandex.ru/metrika/watch.js is already well defended against by Brave.  But, I think thats probably worth thinking through again:

1. limiting to yandex origins seems good; i just wasn't sure i could find a listing of all of them, but if you have them, thats a great idea
2. I'm going to go back to Yandex and see if we can also get the fingerprinting removed

PR #495, opened 2020-11-03
Update new filenames for Brave lists

ryanbr:
This initial commit will split the android/ios filters from unbreak. Just empty text files currently.

Leaving,  `brave-android-specific.txt`, `brave-ios-specific.txt` and `brave-unbreak.txt`


antonok-edm (approved):

antonok-edm (changes requested):
We can't delete `brave-social` just yet, as it's already in `default.json`

antonok-edm (approved):

PR #509, opened 2020-11-18
add back in local resource blocking

pes10k:
We should continue blocking websites trying to scan local services / resources


ryanbr:
Safe to close @pes10k ? https://github.com/brave/adblock-lists/pull/510  was commited

PR #562, opened 2021-02-08
Fix playback on https://www.cbssports.com/nfl/superbowl/live

ryanbr:
Fixes breakage on `https://www.cbssports.com/nfl/superbowl/live` due to Peter Lowes List.


ryanbr:
Included in Brave until https://github.com/brave/adblock-resources/pull/18 is landed

PR #577, opened 2021-03-04
Tracking snippet, no longer needed in unbreak

ryanbr:
After debugging, this filter is too problematic. We'll remove.


ryanbr:
Not needed, closed.

ryanbr:
Okay, reconfirmed. this filter isn't needed. May cause issues

PR #586, opened 2021-03-29
Fix localhost issue on crust.network

ryanbr:
Was reported via https://community.brave.com/t/specific-filter-whitelist-application/223212

> Crust Apps (a Dapp of Crust Network, you can check this: https://apps.crust.network) basically provides decentralized IPFS Pin services. But when we used Brave to explore Crust Apps, we found that Brave Shield would intercept the connection request of 127.0.0.1 from our Apps, and users need to manually close Brave Shield for normal use.


pes10k:
This will need to be reverted.

PR #603, opened 2021-05-16
Remove ios/android exceptions (Waiting on next Brave Ad Block Updater) Do not commit just yet

ryanbr:
In Easylist we've limited the Adblock fixes to [Desktop for the time being](https://github.com/easylist/easylist/commit/f11fc16105756a99d5e689fa15c8f92645b08e39). 

We won't need these exceptions any more. And will allow us to now debug via `brave://adblock` on android the safely block ads on android mobile (and ios).


ryanbr:
Will wait till `Brave Ad Block Updater - Version: 1.0.938` before rolling this change, to be on the safe side.

mihaiplesa (approved):

PR #610, opened 2021-07-01
Fix kinox.cloud popups from https://github.com/brave/brave-browser/issues/14818

ryanbr:
Fixes https://github.com/brave/brave-browser/issues/14818


ryanbr:
Reverted PR, I don't think this is needed.

PR #645, opened 2021-09-16
add package.json for component updater

pilgrim-brave:
This is needed so we can reference the repository from https://github.com/brave/brave-core-crx-packager so it can pull in `debounce.json`


ryanbr (approved):
Looks good!

PR #680, opened 2021-11-21
Remove consent domains, causes noscroll issues on some sites

ryanbr:
Removing the consent domains, will be rolled out in Easyprivacy. Blocking these domains can cause no scroll issues on some sites.


pes10k:
Ryan could we still block these, and then remove the noscroll issues with scriptlets?

PR #728, opened 2021-12-14
Add google.com debounce

ryanbr:
Fixes debounce `www.google.com`

`https://www.google.com/url?q=https://t.17track.net/%23nums%3D2222062731&source=gmail&ust=2239542218716022&usg=A12Vaw22aVcgCnimxpL3T22Gs--w`

### DIFF
--- a/brave-lists/debounce.json
+++ b/brave-lists/debounce.json
@@ -161,7 +161,8 @@
   },
   {
     "include": [
-      "*://*.youtube.com/redirect*"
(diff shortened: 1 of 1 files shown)


pes10k:
@ryanbr a note on this one, this is the same URL google uses for safebrowsing checking.  I think the plan is to only debounce this URL if the user has safebrowsing disabled, so might be best to back this out of here and open an issue around it and we can sort things out there

PR #729, opened 2021-12-14
Revert https://github.com/brave/adblock-lists/pull/728

ryanbr:
Reverted, as per https://github.com/brave/adblock-lists/pull/728#issuecomment-993144404


pes10k:
Thanks @ryanbr . If you can create the issue (in brave-browser), you, I and @pilgrim-brave can work together to figure out the best way forward.

PR #683, opened 2021-11-23
Update brave-sugarcoat.txt

ShivanKaul:
The previous sugarcoat list looks out of date, I'll ask Stephen Donner to confirm this one https://github.com/brave/brave-browser/issues/18542#issuecomment-975628033

PR #740, opened 2021-12-28
Use only the 2 smallest sugarcoat resources

antonok-edm:
For initial rollout of sugarcoat resources shipped from the CRX packager, we want to keep the resource size as small as possible while still demonstrating that the system works correctly.

This PR reduces the list of resources to just the smallest 2 (each 68060 bytes on disk when stripped of randomized padding).

### DIFF
--- a/brave-lists/brave-sugarcoat.txt
+++ b/brave-lists/brave-sugarcoat.txt
@@ -1,10 +1,2 @@
-||googletagservices.com/tag/js/gpt.js$script,important,domain=vimeo.com,redirect-url=https://pcdn.brave.com/sugarcoat/async-sugarcoat-83c6e054eb06b453d56c97b9ddd181d6.js
-||z.moatads.com/reutersheader194883552024/moatheader.js$script,important,domain=reuters.com,redirect-url=https://pcdn.brave.com/sugarcoat/async-sugarcoat-7754008b0da848856cddb20094976335.js
-||sourcepointcmp.bloomberg.com/ccpa.js$script,important,domain=bloomberg.com,redirect-url=https://pcdn.brave.com/sugarcoat/async-sugarcoat-1e156c8652627e444cc63b54d4de0ffb.js
-||quantcast.mgr.consensu.org/tcfv2/23/cmp2.js$script,important,domain=researchgate.net,redirect-url=https://pcdn.brave.com/sugarcoat/async-sugarcoat-1290d064f0a1f106f4c14d0dbd0aca60.js
-||assets.adobedtm.com/launch-ENb97d7f9d2d4b4720ac9782a711994995.min.js$script,important,domain=amazon.jobs,redirect-url=https://pcdn.brave.com/sugarcoat/async-sugarcoat-dd32890a88dc14e37c80777ca67c1e2e.js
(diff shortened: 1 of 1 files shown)


pes10k (approved):

pes10k:
lgtm.  @ShivanKaul is still on PTO for a bit so i think might be best to merge w/o waiting for the additional review, if his review isn't strictly needed

brave-lists/brave-sugarcoat.txt
```diff
@@ -1,10 +1,2 @@
...
-||assets.adobedtm.com/extensions/EPbde2f7ca14e540399dcc1f8208860b7b/AppMeasurement_Module_AudienceManagement.min.js$script,important,domain=bmw.com.au,redirect-url=https://pcdn.brave.com/sugarcoat/async-sugarcoat-4c15a27239807d1629304fcae3cf6ce3.js
-||assets.adobedtm.com/extensions/EPbde2f7ca14e540399dcc1f8208860b7b/AppMeasurement_Module_AudienceManagement.min.js$script,important,domain=virginmobile.ca,redirect-url=https://pcdn.brave.com/sugarcoat/async-sugarcoat-d357b361eae50395667a4c91732192fd.js
-||quantcast.mgr.consensu.org/tcfv2/cmp2.js?referer=einthusan.tv$script,important,domain=einthusan.tv,redirect-url=https://pcdn.brave.com/sugarcoat/async-sugarcoat-968076ccecd06aaae1ffcb1e3d26399f.js
-||googletagservices.com/tag/js/gpt.js$script,important,domain=downdetector.com,redirect-url=https://pcdn.brave.com/sugarcoat/async-sugarcoat-a8badcefe5197e12ea29f0f5d0db1cc9.js
+||assets.adobedtm.com/extensions/EPbde2f7ca14e540399dcc1f8208860b7b/AppMeasurement_Module_AudienceManagement.min.js$script,important,domain=bmw.com.au,redirect=async-sugarcoat-4c15a27239807d1629304fcae3cf6ce3.js
```
ShivanKaul, line 1: I believe the redirect filter option does not take in the file extension (.js), so the filter option would just be `redirect=async-sugarcoat-4c15a27239807d1629304fcae3cf6ce3`. However, I think these 2 rules are obsolete: https://github.com/brave/brave-browser/issues/18542#issuecomment-975628033

I have a local PR that adds a few new ones (10 rules), I can update that to be just 2 rules. I thought we were waiting for https://github.com/brave/brave-core/pull/10994 to be closed?

antonok-edm: > I believe the redirect filter option does not take in the file extension (.js)

For full compatibility with uBO, `redirect` accepts the version of the redirect resource with the `.js` extension or without. Although you're correct in that it's usually omitted.

> I thought we were waiting for brave/brave-core#10994 to be closed?

That PR is still being reviewed; we'd like to get some demo working sooner so as small of a subset as possible would be ideal.

> I have a local PR that adds a few new ones (10 rules), I can update that to be just 2 rules.

That'd be great!

ShivanKaul: I updated https://github.com/brave/adblock-lists/pull/683/files, and we would also need https://github.com/brave/adblock-resources/pull/51

PR #743, opened 2022-01-04
Fixes Anti-adblock on havoc-os.com

ryanbr:
Fixes BlockAdblock on `havoc-os.com`

Countered in uBO via redirect-rule `||pagead2.googlesyndication.com/pagead/js/adsbygoogle.js$script,redirect-rule=googlesyndication_adsbygoogle.js:5`


pes10k:
@ryanbr can we create another, separate file for anti-adblock rules, and move this rule (and the other anti-adblock rules) into their own file? 

Wouldn't change anything Brave-user side, but would be useful for us organizationally, and maybe useful for other projects interested in using the list

ryanbr:
Most of the anti-adblock stuff is related to redirect-rule function. If we add support for this, should eliminate this and others showing anti-adblock via redirect-rules.  https://github.com/brave/adblock-rust/issues/169  @antonok-edm near support for it?

Regarding anti-adblock that I caught affecting Brave and uBO, I notify uBO devs via Slack to land.

antonok-edm:
`redirect-rule` should be supported already in-browser if it's sent as part of a DAT - I just need to update `adblock-rust` to parse those rules and include them. Shouldn't be too much effort, as long as I can squeeze in some time for it. But once implemented, we could be ready to ship it without waiting for clients to update.

PR #744, opened 2022-01-06
add additional localhost alias rules

pes10k:
Added additional filter list rules to handle additional localhost aliases (along with a handy-to-me script to generate all the permutations of different localhost aliases)

Partially addresses https://github.com/brave/brave-browser/issues/9860#issuecomment-1006216069

### DIFF
--- a/tools/gen_localhost_rules.py
+++ b/tools/gen_localhost_rules.py
@@ -1,0 +1,35 @@
+#!/usr/bin/env python3
+
+"""Silly helper script to generate all the permutations of localhost aliases."""
+
+LOCALHOST_ALIASES = [
+    "localhost",
+    "127.0.0.1",
+    "[::1]",
+    "[::ffff:7f00:1]",
+]
+
+NON_ROUTEABLE_ALIASES = [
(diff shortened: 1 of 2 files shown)


fmarier (commented):
It looks like Chrome canonicalizes `[::ffff:0.0.0.0]` as `[::ffff:0:0]`, at least from what I can see in the devtools.

tools/gen_localhost_rules.py
```diff
@@ -0,0 +1,35 @@
...
+    "localhost",
+    "127.0.0.1",
+    "[::1]",
+    "[::ffff:7f00:1]",
+]
+
+NON_ROUTEABLE_ALIASES = [
+    "0.0.0.0",
+    "[::]",
+    "[::ffff:0.0]",
```
fmarier, line 15: `[::ffff:0:0]` (colon instead of dot)

fmarier (changes requested):

tools/gen_localhost_rules.py
pes10k: done

PR #750, opened 2022-01-10
Add brave-etld-exceptions list

antonok-edm:
Towards closing https://github.com/brave/brave-browser/issues/20426 - we'd like to make a new list with rules duplicated from EasyList, EasyPrivacy, etc. that can block first-party requests in scenarios where we have high confidence that the blocking will not cause breakage.

This PR adds that empty list. We can add rules to it over time, and users who are interested in "beta-testing" will be able to enable it as a custom subscription list in brave://adblock by pointing to the raw text URL.


antonok-edm:
Turns out we're already using https://github.com/brave/adblock-lists/blob/master/brave-lists/brave-firstparty.txt for this purpose!

PR #757, opened 2022-01-22
Fix anti-adblock message on topspeed.com

ryanbr:


ryanbr:
Cancelling this PR for the moment, will work with uBO authors

PR #803, opened 2022-03-23
Fix ad clicks on google.com

ryanbr:
Peter Lowes list includes blocks on:

```
127.0.0.1 googleadservices.com
127.0.0.1 doubleclick.net
```

If user intentionally clicks on a Advert on google.com, they'll be presented with an error message. In Easylist we're blocking both of these servers in a 3p context, when it's 1p it'll generate this error.  This issue was previously addressed in IOS and has the same affect on Desktop: https://github.com/brave/adblock-lists/pull/584

Doubleclick and googleadservices.com scripts are still blocked after this change, just addresses people intentionally clicking on ads with Shields up (on standard).

image (15)


ryanbr:
Not applicable for the test case, closing for the time being. Related to a different issue

PR #805, opened 2022-03-24
Remove googleadservices.com debounce

ryanbr:
Removing this debounce, reportedly causing issues when clicking on ad links. 

image (4)

### DIFF
--- a/brave-lists/debounce.json
+++ b/brave-lists/debounce.json
@@ -229,15 +229,6 @@
   },
   {
     "include": [
-      "*://*.googleadservices.com/*"
-    ],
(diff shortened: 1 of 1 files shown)


fmarier:
@ryanbr Do you remember whether we were able to determine why this one broke?

From the screenshot, it looks like a pretty straightforward one to debounce (at least the first hop to `clickserver.dartsearch.net`). Was the the interaction with a second debounce rule that broke things?

ryanbr:
@fmarier https://bravesoftware.slack.com/archives/C2FQMN4AD/p1644940249798829 has most of details

PR #806, opened 2022-03-24
uBO wildcard fixes for kissanime/kissasian

ryanbr:
Fixes popups and scripts used on various kissanime, kissasian mirrors. 

Imported from the latest uBO filters.txt and mirrors available on kissanime/kissasian. And resolve these issues;

kissanie


ryanbr:
Also landed upstream for ubO (missing snippet), https://github.com/brave/adblock-lists/pull/806

PR #878, opened 2022-06-06
Add rfvk.net debounce

ryanbr:
Adding debounce `https://adorama.rfvk.net/c/1305166/51926/1036?u=https%3A%2F%2Fwww.adorama.com%2Fsomve1us1blk.html&subid1=dealmaster060422`

@fmarier

### DIFF
--- a/brave-lists/debounce.json
+++ b/brave-lists/debounce.json
@@ -255,6 +255,7 @@
       "*://*.uqhv.net/c/*",
       "*://*.vzew.net/c/*",
       "*://*.qflm.net/c/*",
+      "*://*.rfvk.net/c/*",
       "*://redirect.viglink.com/*"
     ],
     "exclude": [


fmarier (commented):
One question though, are there other examples where the hostname is not `adorama`?

If not, then we should remove the wildcard at the beginning of the hostname.

ryanbr:
Currently its just related to the adorama domain for this commit, but I use a wildcard just to avoid future updating of urls. Most of [these urls](https://github.com/brave/adblock-lists/blob/master/brave-lists/debounce.json#L212) have the same patterns. None of these domains host anything or used for anything but for bouncing urls off shopping sites.

fmarier:
Now that I've reviewed all of the Brave-added rules, I see what you mean. It looks like this particular adtech company is using disposable domains and put the client name as a subdomain, but it's not even required:
```
$ curl --head -L 'https://foobar.rfvk.net/c/1305166/51926/1036?u=https%3A%2F%2Fwww.adorama.com%2Fsomve1us1blk.html&subid1=dealmaster060422'
HTTP/2 301 
...
location: https://adorama.rfvk.net/c/1305166/51926/1036?u=https%3A%2F%2Fwww.adorama.com%2Fsomve1us1blk.html&subid1=dealmaster060422&level=1&brwsr=bot&brwsrsig=3jr0xZWggwVt1kMSob1zbxttU6%3ATTT

HTTP/2 301 
...
location: https://www.adorama.com/somve1us1blk.html?sterm=&kbid=915043

```

PR #889, opened 2022-06-23
Add test entries to the debounce list.

fmarier:
These dummy entries can be used in tests since they will be stable.

### DIFF
--- a/brave-lists/debounce.json
+++ b/brave-lists/debounce.json
@@ -392,5 +392,33 @@
     ],
     "action": "base64,redirect",
     "param": "a"
+  },
+  {
+    "include": [
+      "https://fmarier.github.io/brave-testing/debouncer/redirect.html?*"
+    ],
+    "exclude": [
+    ],
+    "action": "redirect",
+    "param": "brave_testing"
+  },
+  {
+    "include": [
+      "https://fmarier.github.io/brave-testing/debouncer/redirect.html?*"
+    ],
+    "exclude": [
+    ],
+    "action": "base64,redirect",
+    "param": "brave_testing_base64"
+  },
+  {
+    "include": [
+      "https://fmarier.github.io/brave-testing/debouncer/*"
+    ],
+    "exclude": [
+    ],
+    "prepend_scheme": "https",
+    "action": "regex-path",
+    "param": "^/brave-testing/debouncer/(.*);.html$"
   }
 ]


pilgrim-brave (changes requested):
These should be Brave-controlled domains, yes?

fmarier:
> These should be Brave-controlled domains, yes?

If @pes10k has time to help, sure, but I don't have access to the test page he runs.

Since the features have landed / are landing and QA has no way to test them, this is a way to unblock these PRs. We can always update the list later to point to another domain.

brave-lists/debounce.json
```diff
@@ -392,5 +392,34 @@
...
+    "exclude": [
+    ],
+    "action": "base64,redirect",
+    "param": "brave_testing_base64"
+  },
+  {
+    "include": [
+      "https://fmarier.github.io/brave-testing/debouncer/*"
+    ],
+    "pref": "brave.de_amp.enabled",
```
pes10k, line 418: This part doesn't seem correct to me.  We want to allow regex rules in general, independent of the de-amping feature being enabled.  There are just certain specific debouncing rules we want to enable only when the DeAMPing feature is enabled

pes10k (changes requested):

brave-lists/debounce.json
fmarier: Good point. I copy/pasted that from Shivan's issue, but we don't need it for this.

pes10k:
okay test cases are here: https://dev-pages.brave.software/navigation-tracking/debouncing.html

pes10k (approved):

pes10k:
I agree with @pilgrim-brave though, it would be good to get these moved to Brave owned domains.  Maybe we could do that as a follow up ticket

fmarier:
Yes, if we can find another Brave domain we can use, then we should host the "bouncers" there.

pes10k:
here are the three domains we use for tests.

these two are the same, and serve only static content (they mirror each other to test for things like storage partitioning).  I have access to the S3 buckets that serve these pages
http://dev-pages.brave.software/
https://dev-pages.bravesoftware.com/

There is also a server at mq6xb08mii.execute-api.us-west-2.amazonaws.com/development we use for these tests that has dynamic code on it.  Right now its only used for the websockets test, but might be a good place to put more dynamic code (like debouncing stuff).  I do not have access to this server though, would need devops

fmarier:
Follow-up issue: https://github.com/brave/adblock-lists/issues/890

PR #897, opened 2022-06-24
Debounce clixgalore.com correctly

fmarier:
Clixgalore uses two different query string parameters, and it also omits the scheme from the destination URLs.

Sample URL: http://www.clixgalore.com/Lead.aspx?BID=188989&AfID=82776&AdID=16214&LP=www.elitemate.com%2FmyJsp%2Fjoin.jsp%3Fid%3Dclixg%26path%3DPG%26pageid%3D339

Note that this uses the new "prepend_scheme" option which recently landed. This was not available when the rule was originally added in https://github.com/brave/adblock-lists/pull/770.

### DIFF
--- a/brave-lists/debounce.json
+++ b/brave-lists/debounce.json
@@ -65,12 +65,23 @@
   },
   {
     "include": [
-      "*://*.clixgalore.com/Lead.aspx*"
+      "*://www.clixgalore.com/Lead.aspx*"
     ],
     "exclude": [
     ],
+    "prepend_scheme": "http",
     "action": "redirect",
     "param": "AffDirectURL"
+  },
+  {
+    "include": [
+      "*://www.clixgalore.com/Lead.aspx*"
+    ],
+    "exclude": [
+    ],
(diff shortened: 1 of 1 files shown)


fmarier:
That's a feature that @ShivanKaul just added. Basically, if the parameter only contains a hostname (in the example above we can see `www.elitemate.com` instead of `https://www.elitemate.com` or `http://www.elitemate.com`) then the debounce isn't going to work because it's missing the scheme and isn't a valid URL.

So `prepend_scheme` allows us to specify that a rule contains such an incomplete URL and instructs the debouncer to tack on a specific scheme (`http` or `https`) prior to debouncing. If there's already a scheme in the destination URL, then adding another one is going to fail, so this is just for the case where it's missing from the destination URL.

Shivan will be updating the docs soon for this new option, as well as for the new rule type he added (`regexp-path`).

fmarier:
https://github.com/brave/adblock-lists/pull/898 is another example of rules that need the scheme to be prepended.

PR #907, opened 2022-06-27
Remove domain without a sample URL

fmarier:
This was added in https://github.com/brave/adblock-lists/pull/706
but a sample URL wasn't included.

I searched for an example on Google and archive.org and couldn't
find one and so we should remove and re-add later if we find one.


fmarier:
Awesome, you still had a copy of it. I added it to #706 so we have a record of this.

PR #911, opened 2022-07-04
Add cc.techbargains.com debounce

ryanbr:
From site: `https://www.techbargains.com/deals/lg-oled-c1-4k-tv`

Debouncing `https://cc.techbargains.com/v1/otc/06LJoHfGQoCuYdVtZVoTdgY?merchant=03yqSJWwdU65jruBzT9LVdv&url=https%3A%2F%2Fwww.buydig.com%2Fshop%2Fproduct%2FLGOLED77C1PUB&i=deal_description&e=deal_tile&p=1&m=deals_list&t=deal_page`

### DIFF
--- a/brave-lists/debounce.json
+++ b/brave-lists/debounce.json
@@ -25,7 +25,8 @@
       "*://go.redirectingat.com/*",
       "*://x.chip.de/linktrack/button/*",
       "*://*.mailchimp.com/mctx/clicks?*",
-      "*://*.tradedoubler.com/click?*"
+      "*://*.tradedoubler.com/click?*",
+      "*://cc.techbargains.com/*"
     ],
     "exclude": [
     ],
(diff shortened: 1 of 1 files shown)


brave-lists/debounce.json
```diff
@@ -25,7 +25,8 @@
       "*://go.redirectingat.com/*",
       "*://x.chip.de/linktrack/button/*",
       "*://*.mailchimp.com/mctx/clicks?*",
-      "*://*.tradedoubler.com/click?*"
+      "*://*.tradedoubler.com/click?*",
+      "*://cc.techbargains.com/*"
```
fmarier, line 29: I would include this part of the path as well: `/v1/otc/`, in other words:
```
      "*://cc.techbargains.com/v1/otc/*"
```
If they change to another version of their API (e.g. `/v2/otc/...`) then we may need to revise the rule type or the name of the parameter.

fmarier (changes requested):

fmarier:
Also, here's a trick I use to minimize the size of the diff.

Since adding to the end means you need to add a comma to the last entry and then add the new entry afterwards, it ends up touching two lines. I usually try to add it just before the last entry (just before the first would work as well) and so the diff is a single line.

A nice side-effect of this, is that it's easier to look through the git history (e.g. using the `git blame` functionality) if lines don't get modified for "unnecessary" reasons.

PR #913, opened 2022-07-05
Fixes https://signin.quicken.com/

ryanbr:
Removed `53.com` dube domain. 

Added quicken.com (Reported https://community.brave.com/t/quicken-com-sign-in-page-always-hangs/409691/2)


ryanbr:
Also added `spectrum.net`  (Same threatmetrix script reported in the webcompat reports)

PR #940, opened 2022-08-08
Debounce AMP cache URLs

ShivanKaul:
Using the new `regex-path` action, debounce AMP cache URLs to the link embedded in the cache URL.

These AMP cache URLs look like https://www-nytimes-com.cdn.ampproject.org/c/s/www.nytimes.com/wirecutter/blog/burner-browser-to-hide-internet-searches/amp/
The regex is in the param: it captures the embedded URL after the `/c/s` in the path.
Add an `https` scheme to the beginning, because the embedded link doesn't have a scheme.
Only apply debouncing rule if the `de_amp` user pref is turned on.

### DIFF
--- a/brave-lists/debounce.json
+++ b/brave-lists/debounce.json
@@ -364,5 +364,16 @@
     "prepend_scheme": "https",
     "action": "regex-path",
     "param": "^/navigation-tracking/(.*);.html$"
+  },
+  {
+    "include": [
+      "*://*.cdn.ampproject.org/c/s/*"
+    ],
+    "pref": "brave.de_amp.enabled",
+    "exclude": [
+    ],
+    "prepend_scheme": "https",
+    "action": "regex-path",
+    "param": "^/c/s/(.*)$"
   }
 ]
--- a/brave-lists/debounce.json
+++ b/brave-lists/debounce.json
@@ -336,6 +336,17 @@
   },
   {
     "include": [
+      "*://*.cdn.ampproject.org/c/s/*"
+    ],
+    "pref": "brave.de_amp.enabled",
+    "exclude": [
+    ],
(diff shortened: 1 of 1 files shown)


pes10k:
This is terrific @ShivanKaul ! 

Just a general question for my own education though, do you know what the `/c/s` path entries are for?  Are there other paths on the AMP cache we should handle differently?  (this isn't a blocking issue of course, just curious if you came across this info when working on this feature)

pes10k (approved):
Terrific!

brave-lists/debounce.json
```diff
@@ -364,5 +364,16 @@
     "prepend_scheme": "https",
     "action": "regex-path",
     "param": "^/navigation-tracking/(.*);.html$"
+  },
```
pes10k, line 367: here is a very very minor request, but could you move the new rules _before_ the internal / brave testing ones?  Its a super minor nit kind of thing, but would just be nice to not have the brave-QA-testing rules in the middle of "real world" rules

fmarier (approved):
I agree with @pes10k . Let's keep the test entries at the end of the file.

ShivanKaul:
@pes10k @fmarier updated to keep the test rules at the bottom of the file.

The `/c/` stands for **content** and is used for standalone AMP pages. It could also be `/i/` for **images** or `/r/` for **resources**, but debouncing should only happen for documents IIUC. The /s/ is optional and stands for **secure** i.e. HTTPS (source: https://developers.google.com/amp/cache/overview).

I haven't seen any of these in the wild, but I suppose if there's no `/s/` then we could prefix plain-HTTP (`prepend_scheme: http` instead of `https` like we have now) to the destination URL.

PR #948, opened 2022-08-24
Extend test harness for new JSON list

fmarier:
The new list will be populated in a later PR, but I want to make sure tests are in place first.


kim0 (approved):
Thanks!
I don't know if we'll be adding even more json files in the future. If we will, maybe we should consider [glob'ing](https://docs.python.org/3/library/glob.html) making the tool able to handle an arbitrary number of files?

fmarier:
> If we will, maybe we should consider [glob'ing](https://docs.python.org/3/library/glob.html) making the tool able to handle an arbitrary number of files?

That's a good idea. I'm not sure we'll be adding new ones (this `copy-urls.json` one is temporary since the long-term plan is to move to an externally-managed list) but we should definitely do that if there's a third one :)

PR #961, opened 2022-09-04
Add email subscriptions tracking

ryanbr:
From: https://github.com/AdguardTeam/AdguardFilters/blob/master/TrackParamFilter/sections/general_url.txt

**! https://github.com/AdguardTeam/AdguardFilters/issues/89869 ! email subscriptions tracking**
```
$removeparam=bsft_clkid
$removeparam=bsft_eid
$removeparam=bsft_mid
$removeparam=bsft_uid
$removeparam=bsft_aaid
$removeparam=bsft_ek
```

**! Matomo/Piwik** 
```
$removeparam=mtm_campaign
$removeparam=mtm_cid
$removeparam=mtm_content
$removeparam=mtm_group
$removeparam=mtm_keyword
$removeparam=mtm_medium
$removeparam=mtm_placement
$removeparam=mtm_source
$removeparam=pk_campaign
$removeparam=pk_medium
$removeparam=pk_source
```
**! Branch.io ( https://stackoverflow.com/questions/37882810/what-is-meaning-of-branch-match-id )**
`$removeparam=_branch_match_id`

**! Valuecommerce ( https://help.valuecommerce.ne.jp/mer/dl/ITP2.0_for_iTAG.pdf )**
`$removeparam=vc_lpp`

**! https://github.com/brave/brave-browser/issues/17507**
```
$removeparam=ml_subscriber
$removeparam=ml_subscriber_hash
```

### DIFF
--- a/brave-lists/clean-urls.json
+++ b/brave-lists/clean-urls.json
@@ -60,7 +60,18 @@
             "bsft_mid",
             "bsft_uid",
             "bsft_aaid",
-            "bsft_ek"
+            "bsft_ek",
+            "mtm_campaign",
+            "mtm_cid",
+            "mtm_content",
+            "mtm_group",
+            "mtm_keyword",
+            "mtm_medium",
+            "mtm_placement",
+            "mtm_source",
+            "pk_campaign",
+            "pk_medium",
+            "pk_source"
         ]
     },
     {
--- a/brave-lists/clean-urls.json
+++ b/brave-lists/clean-urls.json
@@ -73,7 +73,9 @@
             "pk_medium",
             "pk_source",
             "_branch_match_id",
-            "vc_lpp"
+            "vc_lpp",
+            "ml_subscriber",
+            "ml_subscriber_hash"
         ]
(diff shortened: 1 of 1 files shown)


fmarier:
@ryanbr If you see any entries that come from Brave (`ml_subscriber` and `ml_subscriber_hash`), then they don't need to be added to the clean-url list since they'll be removed automatically earlier. 

They're not wrong, nor will they cause problems, they're just not necessary.

pes10k:
just a thought, but you might want to leave them in the clean URL list (even if they're duplicates) is that we don't strip query parameters on same-site navigations, so there are at least possible cases where the query params are in the URL, but users still might want them removed.

fmarier:
Good point. I don't think it's going to matter in this particular case since they're email newsletter trackers (therefore will be cross-site), but it may in other cases.

fmarier:
PR to add the bsft ones to the main query filter: https://github.com/brave/brave-core/pull/15008

PR #965, opened 2022-09-07
GitHub Linguist support for adblock filters

scripthunter7:
GitHub has been supporting adblock syntax highlighting since September 5th. This PR will help you turn it on.

Further informations:
- Syntax highlighter repository: https://github.com/ameshkov/VscodeAdblockSyntax
- How Linguist works: https://github.com/github/linguist/blob/master/docs/how-linguist-works.md
- Linguist overrides: https://github.com/github/linguist/blob/master/docs/overrides.md
- Linguist commit: https://github.com/github/linguist/commit/e78ef71af3600f96b9a40a06511ebbe3797e7401

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+brave-*.txt linguist-language=AdBlock linguist-detectable
+coin-*.txt linguist-language=AdBlock linguist-detectable
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -1,3 +1,4 @@
+[Adblock Plus 2.0; uBlock Origin]
 ||ntv.io^$third-party
 @@||adm.fwmrm.net^*/AdManager.js$domain=msnbc.com|sky.com|cnbc.com
 ||novately.com^$third-party
--- a/coin-miners.txt
+++ b/coin-miners.txt
@@ -1,4 +1,4 @@
-[Adblock Plus 2.0]
+[Adblock Plus 2.0; uBlock Origin]
 !-------------------------------------------------------------------------------
 ! Title: NoCoin Filter List
 ! Expires: 2 days (update frequency)


.gitattributes
```diff
@@ -0,0 +1,2 @@
+brave-*.txt linguist-language=AdBlock linguist-detectable
+coin-*.txt linguist-language=AdBlock linguist-detectable
```
antonok-edm, line 2: @ryanbr @ShivanKaul can we move `debounce.json` out of `brave-lists`, and then move `brave-unbreak.txt` and `coin-miners.txt` into `brave-lists`? That should make the organization here a bit simpler.

brave-unbreak.txt
```diff
@@ -1,3 +1,4 @@
+[Adblock Plus 2.0; uBlock Origin]
```
antonok-edm, line 1: do we need this on the entries under `brave-lists/*` as well?

antonok-edm (commented):
@scripthunter7 thanks for the PR! It'll be nice to have syntax highlighting in this repo.

.gitattributes
ShivanKaul: In favour of moving `brave-unbreak.txt` and `coin-miners.txt` into `brave-lists`, but not sure where else `debounce.json` could live because it is technically a `brave-list`. Maybe we could have a sub-folder here with all the true adblock lists?

brave-unbreak.txt
scripthunter7: @antonok-edm Actually, the `.gitattributes` file would be enough for this, since it overrides the language for the given patterns, and this behavior is useful for handling sections, eg. `brave-lists/*.txt` in this repo. However, it is advisable to specify the adblock agent for complete lists. It is important to know that one of the conditions of Linguist support is the fulfillment of the [**popularity threshold**](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-an-extension-to-a-language), and Linguist only examines the adblock agent with a heuristic. If the use of the adblock agent is reduced (because it is replaced by a gitattributes file), then the heuristics support may cease in the future. However, this heuristic is the only way to clearly distinguish a legit txt file from an adblock list, so the agent is one of the cornerstones of adblock language support.

Therefore, if possible, use the agent for the complete list, but you can use .gitattributes for partial lists (or call it sections or something). Anyway, you can also use the agent for sections, that's also good.

antonok-edm: @scripthunter7 makes sense, thanks for the detailed answer!

.gitattributes
antonok-edm: That sounds reasonable. I'll note this down in a new issue and we can handle it in a followup.

antonok-edm (approved):

PR #984, opened 2022-10-09
Fixes Anti-brave checks on https://zoro.to/

ryanbr:
Fixes anti-Brave checks. Unable to playback video.

zoroto


SangeethSujith:
This issue is still there for zoro.to

PR #1035, opened 2022-12-05
Remove params sync from brave-core

ryanbr:
With merging of https://github.com/brave/adblock-rust/pull/235 We can move away from hardcoded params.

Syncing with hard-coded params from https://github.com/brave/brave-core/blob/master/browser/net/brave_query_filter.cc

`Included comments`
(Generic filters): https://github.com/brave/brave-core/blob/master/browser/net/brave_query_filter.cc#L21  
(Site Sepecific): https://github.com/brave/brave-core/blob/master/browser/net/brave_query_filter.cc#L68

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -915,6 +915,9 @@ $removeparam=guce_referrer
 $removeparam=guce_referrer_sig
 ! https://github.com/brave/brave-browser/issues/26295
 $removeparam=vgo_ee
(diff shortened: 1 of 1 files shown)


ryanbr:
cc: @fmarier

fmarier:
This PR is now out-of-sync with the list that ships with the browser. Let's resurrect it once this use-case is supported in brave-core.

PR #1049, opened 2022-12-17
fix rbs.ta36.com anti-adb

Yuki2718:
This can not be addressed in AG Japanese as already addressed by AG Base in accordance with AG policy.

ta36

I have almost no idea where to put the rule, tell me if it's wrong place. You'll need the same fix for all or most of CHP sites as uBf relies on $redirect-rule not supported by Brave to counter them.


Yuki2718:
9ketsuki

PR #1067, opened 2023-01-25
Add streaklinks debounce

ryanbr:
Was discussed in https://old.reddit.com/r/uBlockOrigin/comments/10kryo7/streaklinkscom/

From the example:

`https://streaklinks.com/BQ_bCtlcnH91IqmN1QFG0nUm/https%3A%2F%2Fwww.congress.gov%2Fbill%2F117th-congress%2Fhouse-bill%2F6707`

### DIFF
--- a/brave-lists/debounce.json
+++ b/brave-lists/debounce.json
@@ -382,7 +382,7 @@
     "exclude": [
     ],
     "action": "regex-path",
-    "param": "^/([^/]+)/s(/.*)$"
+    "param": "^/[^/]+/(.*)$"
   },
   {
     "include": [
--- a/brave-lists/debounce.json
+++ b/brave-lists/debounce.json
@@ -381,9 +381,8 @@
     ],
     "exclude": [
     ],
-    "prepend_scheme": "https",
     "action": "regex-path",
-    "param": "^/streaklinks.com/(.*)$"
+    "param": "^\/[^\/]+\/(.*)$"
   },
   {
     "include": [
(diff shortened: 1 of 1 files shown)


brave-lists/debounce.json
```diff
@@ -375,6 +375,16 @@
...
   },
+  {
+    "include": [
+      "*://streaklinks.com/*"
+    ],
+    "exclude": [
+    ],
+    "prepend_scheme": "https",
+    "action": "regex-path",
+    "param": "^/streaklinks.com/(.*)$"
```
fmarier, line 386: This is not going to work because it needs to skip the `BQ_bCtlcnH91IqmN1QFG0nUm` bit. Also, the regular expression is just over the path of the URL. Right now, this is looking for the hostname too.

So I think (I haven't tested it), we need something like: `^/[^/]+/(.*)$`

fmarier (changes requested):

brave-lists/debounce.json
ryanbr: Should this be `^\/[^\/]+\/(.*)$` rather than `^/[^/]+/(.*)$`  @ShivanKaul ?  Not too familar with the regex syntax used here

fmarier: No need to escape the slashes, see this example here: https://github.com/brave/adblock-lists/blob/ab585082db75a2afd6e88874559c0971fb6faaed/brave-lists/debounce.json#L387

fmarier (approved):

PR #1080, opened 2023-02-17
Medium email campaign source qs remove

### LINKED ISSUE #1079: Medium Digest Email - remove source querystring
Currently Medium (medium.com) sends out digest emails containing links to articles, but the links have query-string that uniquely ties to registered email account.

```
https://medium.com/@simonholdorf/101-tips-to-make-you-a-better-developer-1fbb85d7eddc?source=email-d3383424d595-1676573507630-digest.reader-5517fd7b58a6-1fbb85d7eddc----0-71------------------03f2eee0_f86b_4387_a628_e17bb773f5e1-27
```

Would like to remove `source` as a querystring

legopin:
FIX #1079 

Removes email account identifying link in urls accessed from Medium Email Digest article links

### DIFF
--- a/brave-lists/clean-urls.json
+++ b/brave-lists/clean-urls.json
@@ -213,5 +213,16 @@
             "brave_testing1",
             "brave_testing2"
         ]
-    }
+    },
+    {
+        "include": [
+            "*://medium.com/*"
+        ],
+        "exclude": [
+
+        ],
+        "params": [
+            "source"
+        ]
+    },
 ]
--- a/brave-lists/clean-urls.json
+++ b/brave-lists/clean-urls.json
@@ -148,6 +148,17 @@
     },
     {
         "include": [
+            "*://medium.com/*"
+        ],
+        "exclude": [
+
+        ],
+        "params": [
+            "source"
+        ]
+    },
+    {
+        "include": [
             "*://*/*"
         ],
         "exclude": [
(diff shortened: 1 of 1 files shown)


brave-lists/clean-urls.json
```diff
@@ -213,5 +213,16 @@
...
+        "include": [
+            "*://medium.com/*"
+        ],
+        "exclude": [
+
+        ],
+        "params": [
+            "source"
+        ]
+    },
```
fmarier, line 227: This will make the JSON file invalid because trailing commas are not allowed, unfortunately. I suggest simply moving this new block to be just below the YouTube block (around line 147). That means there's no need to add the annoying comma at the end, and we keep the test entry last.

fmarier (changes requested):
Thanks for the new rule! Just one small thing to fix and we can merge.

pes10k:
@fmarier could we add a "is valid JSON check" as a commit hook?

pes10k:
also, seems like we should move this to the query param stripping feature too, and not just the clean copy (if this param is tying things to a specific user)

fmarier:
> @fmarier could we add a "is valid JSON check" as a commit hook?

We already do this, but CI doesn't run by default on PRs submitted by someone outside the `brave` org.

fmarier:
> also, seems like we should move this to the query param stripping feature too, and not just the clean copy (if this param is tying things to a specific user)

Good idea: https://github.com/brave/brave-browser/issues/28649

PR #1097, opened 2023-03-01
Add Brave Ads query decorations

fmarier:
Slightly edited example URL from an upcoming push notification campaign: https://www.example.com/electronic-widget/?utm_source=atn&utm_campaign=122-cc06eb&utm_medium=native&product=electronic-widget&clickid=500&brave-campaign-id=12e88c57227e&brave-creative-set-id=c586ac2b937&brave-creative-id=c34609-6133

These params are documented here: https://github.com/brave/brave-browser/wiki/Brave-Ads-Reserved-click-through-URL-parameters#reserved-click-through-url-parameters

### DIFF
--- a/brave-lists/clean-urls.json
+++ b/brave-lists/clean-urls.json
@@ -187,6 +187,9 @@
 
         ],
         "params": [
+            "brave-campaign-id",
+            "brave-creative-id",
+            "brave-creative-set-id",
             "ml_subscriber",
             "ml_subscriber_hash",
             "mc_cid",


pes10k:
this looks good to me, but can you say more @fmarier about where these would be showing up in page URLs?

fmarier:
@pes10k Good point, we should have some example URLs.

@iambrianfung Do you have any examples we can share here?

fmarier:
I've added an example URL to the issue description.

iambrianfung (approved):

pes10k:
happy to move the conversation somewhere else if its better but:

1. where is clickid being populated from?  
2. why do we need all three?  seems like 2 of these, and maybe all three, can be described by each other (i.e., are none of the brave-* ids in a one to many or one to one relationship with each other?)

pes10k:
I think this is fine to merge as is btw, but that we should see if we can get some of these brave values removed from the ads system all togehter, since they seem unnecessary and not helpful for our users

ryanbr (approved):

lukemulks:
> I think this is fine to merge as is btw, but that we should see if we can get some of these brave values removed from the ads system all togehter, since they seem unnecessary and not helpful for our users

Hi @pes10k they're needed as the same creative ID can be added to different creative set IDs with different category targets per set,  from more than one campaign ID from the ad server. 

Having the separate parameters allows us to gauge which combination of creative + category is performing best per campaign. 

Campaigns may be targeting different countries, etc. so having them broken out for each is helpful and needed for customers. 

Fwiw, the brave parameters all went through security and privacy review, and we have public facing documentation for them on github https://github.com/brave/brave-browser/wiki/Brave-Ads-Reserved-click-through-URL-parameters

Hope this helps.

tackley:
> Hi @pes10k they're needed as the same creative ID can be added to different creative set IDs with different category targets per set, from more than one campaign ID from the ad server.

In fact, what's called the `brave-creative-id` is in fact the creative _instance_ id. Creatives are indeed shared across many campaigns, but the _creative instance id_ represents the association of a creative with a particular creative set. So in this case the values are strictly hierarchical:

each `brave-creative-id` (i.e. the creative instance id) is associated with exactly one creative set.
each `brave-creative-set-id` is associated with exactly one campaign.

Both the creative set id and the campaign id are derivable from the creative instance id. However, that integrating that derivation into the different analytics systems used by advertisers is likely a difficult hurdle for advertisers to cross.

PR #1099, opened 2023-03-08
Fixes am3t9s.net debounce

ryanbr:
Added debounce for: `https://dsw.am3t9s.net/c/256585/315836/4837?subId1=FN--&u=https%3A%2F%2Fwww.dsw.com%2Fen%2Fus%2Fproduct%2Fdrew-bon-voyage-sandal%2F518930`

### DIFF
--- a/brave-lists/debounce.json
+++ b/brave-lists/debounce.json
@@ -232,7 +232,8 @@
       "*://*.vzew.net/c/*",
       "*://*.qflm.net/c/*",
       "*://*.rfvk.net/c/*",
-      "*://redirect.viglink.com/*"
+      "*://redirect.viglink.com/*",
+      "*://*.am3t9s.net/*"
     ],
     "exclude": [
     ],


fmarier (approved):
That looks fine.

As far as I can tell though (searching Google and publicwww), they don't use any other subdomains than `dsw` and [Amass](https://github.com/OWASP/Amass) doesn't think so either:
```
$ ./amass enum -d am3t9s.net
dsw.am3t9s.net
www.am3t9s.net
am3t9s.net

OWASP Amass v3.22.0                               https://github.com/OWASP/Amass
--------------------------------------------------------------------------------
3 names discovered - archive: 1, scrape: 1, cert: 1
--------------------------------------------------------------------------------
ASN: 16509 - AMAZON-02 - Amazon.com, Inc.
	52.48.0.0/14      	2    Subdomain Name(s)
	52.16.0.0/15      	2    Subdomain Name(s)
	34.240.0.0/12     	2    Subdomain Name(s)
ASN: 14618 - AMAZON-AES - Amazon.com, Inc.
	52.72.0.0/15      	1    Subdomain Name(s)
	44.192.0.0/11     	2    Subdomain Name(s)
	54.196.0.0/15     	1    Subdomain Name(s)

The enumeration has finished
```
But it does look like the hostname is the client name (dsw.com) and they only use [wildcard TLS certs](https://crt.sh/?q=am3t9s.net) and so it seems like they are wanting to let others CNAME into their domain, so we may as well have the wildcard in our rule too.

PR #1105, opened 2023-03-14
Revert "Remove context tracker on Reddit"

### LINKED ISSUE #1104: Reddit's context parameter being stripped from a clean copied link is breaking website functionality
According to [this commit](https://github.com/brave/adblock-lists/commit/908fed12eb348de53918315147257c70e04c1805) Reddit's `context` parameter is considered a tracker. Is there any real basis for this? Because I believe there has been a fundamental misunderstanding of what that parameter does. The URLs given in the commit message don't make sense as samples (ironically, given the context of how `context` is intended to be used), as they do not link to a deeply nested comment — instead they link directly to a thread, where `context` won't change anything in the output of the page — which isn't really a fair argument for its removal.

`context` accepts an integer that determines how many levels of parent comments are shown for, you guessed it, context to a **single comment**. This means any example URLs must link to a single comment with *at least* 1 parent comment for the functionality to work.

Here's an example to demonstrate; a link to a single comment nested within a comment thread:
* https://old.reddit.com/r/WritingPrompts/comments/11qwxqx/wp_the_demon_queen_lays_on_the_floor_soaked_in/jc6403f/

The user's comment makes reference to confusing the word 'exercise' with 'exorcise', but as a standalone comment it is not clear why this is being pointed out. However, if I add the context parameter like so:
* https://old.reddit.com/r/WritingPrompts/comments/11qwxqx/wp_the_demon_queen_lays_on_the_floor_soaked_in/jc6403f/?context=3
And we visit that URL, you will now see that the full parent comment is displayed, and the specific comment linked is still highlighted below it. This gives the user context to the comment being linked.

To illustrate further, we can adjust the value of `context` in that same thread and look at how it effects output.
Here is a link to a 3rd level comment in the same thread, on its own:
* https://old.reddit.com/r/WritingPrompts/comments/11qwxqx/wp_the_demon_queen_lays_on_the_floor_soaked_in/jc6fnvx/
And a link to that same comment with 1 level of parent comment shown with it:
* https://old.reddit.com/r/WritingPrompts/comments/11qwxqx/wp_the_demon_queen_lays_on_the_floor_soaked_in/jc6fnvx/?context=1
And a link to that same comment this time with 2 levels of parent comments shown with it:
* https://old.reddit.com/r/WritingPrompts/comments/11qwxqx/wp_the_demon_queen_lays_on_the_floor_soaked_in/jc6fnvx/?context=2

I fully understand the intent of clean links and why we desire them, but stripping `context` from reddit URLs I believe is a bad idea as it breaks useful functionality on their website. I don't consider it to be "tracking" a user at all and am honestly very confused as to why it has been considered as such in the first place. Especially considering the length of time reddit has been around on the internet, these types of contextual URLs aren't a secret...

I hope this commit can be looked at and reversed, as I find myself sharing links to specific reddit comments fairly often, and it's annoying having to manually write `?context=3` back in each time.

A couple of additional notes:
1. I used old reddit links however the functionality exists on new reddit too. Simply remove `old.` from the URLs if you wish to confirm.
2. Years ago when reddit used to display context links beneath comments, 3 was always the default value, hence why I used it here even though there weren't three levels of parent comments to show.
3. Apologies for linking a wall of text as an example but I was literally trying to share it with a friend and that's what's led me here!

fmarier:
Fixes #1104.

This reverts commit 908fed12eb348de53918315147257c70e04c1805.

### DIFF
--- a/brave-lists/clean-urls.json
+++ b/brave-lists/clean-urls.json
@@ -170,17 +170,6 @@
     },
     {
         "include": [
-            "*://*.reddit.com/*"
-        ],
-        "exclude": [
-
-        ],
-        "params": [
-            "context"
-        ]
-    },
-    {
-        "include": [
             "*://*/*"
         ],
         "exclude": [


ryanbr (approved):
LGTM

PR #1129, opened 2023-04-03
Add more Amazon params from affiliate links

fmarier:
Sample URL: https://www.amazon.com/dp/B002CVTT4S?tag=affiliate-name&linkCode=osi&th=1&psc=1&language=en_US

I'm not sure what `th` or `psc` do, but they're not on the Adguard list and they're pretty short, so maybe we just leave them out for now.

### DIFF
--- a/brave-lists/clean-urls.json
+++ b/brave-lists/clean-urls.json
@@ -28,6 +28,8 @@
 
         ],
         "params": [
+            "linkCode",
+            "tag",
             "crid",
             "qid",
             "sprefix"


ryanbr (approved):

solso:
This would remove the `&tag=` parameter on Amazon URLs, right? If that's the case, we cannot deploy this.

First, we are using &tag= in the shopping cluster on Brave search. So that would basically break all attribution. 

Second, and not less important, there is no grounds to remove this tag from privacy or from ads/attention perspective. It would be disrupting all content creators (including us) that rely on Amazon attribution.

Please, hold the merge.

pes10k:
Josep, this list is only applied on copy-paste, not navigation. Unless people are copy-pasting the URL out of the Amazon carousel (using the “copy clean url” option) and pasting it into a new tab or something, this change will not affect the Brave search feature. That this is not purely a privacy feature is the reason we don’t apply these protections by default, only when a user specifically asks for it in the drop down menuOn Apr 4, 2023, at 04:06, Josep M. Pujol ***@***.***> wrote:﻿
This would remove the &tag= parameter on Amazon URLs, right? If that's the case, we cannot deploy this.
First, we are using &tag= in the shopping cluster on Brave search. So that would basically break all attribution.
Second, and not less important, there is no grounds to remove this tag from privacy or from ads/attention perspective. It would be disrupting all content creators (including us) that rely on Amazon attribution.
Please, hold the merge.

fmarier:
Yes, like @pes10k says, this is not removed automatically. We only strip user trackers automatically when they meet our [policy](https://github.com/brave/brave-browser/wiki/Query-String-Filter) which specifically excludes coarse-grained campaign-level tracking like this `tag=` parameter. In other words, we never intend to block campaign attribution like the one you're describing @solso .

Instead, this is for when users copy/paste URLs from the URL bar using this option:
Screenshot from 2023-04-04 11-11-03

pes10k:
fwiw, discussed with @solso on slack and we're all good to merge.  I'm going to hit the button now to prevent further confusion

PR #1147, opened 2023-04-26
Fixes popups on bookoffonline.co.jp

ryanbr:
Reported on https://community.brave.com/t/topic/483981 

Blocking the following trackers helps remove the popups.


Yuki2718:
All of them are already in EP, why iOS needs them?

pes10k:
> All of them are already in EP, why iOS needs them?

Because iOS cannot use all of easylist (we're limited in the number of network-level rules we can apply at a time by WKWebView).  We use https://github.com/brave/slim-list-lambda to try and guess which rules will be most useful to users, but adding a rule to our explicit iOS list ensures that the list will be shipped to all users

PR #1156, opened 2023-05-05
Create localhost-permission-allow-list.txt

ShivanKaul:
These domains are from https://docs.google.com/spreadsheets/d/1-lK-bY8NeOGsnwS__joZ7-GJLy-C8nk80tINOWalzhc/edit#gid=0

I've left these as eTLD+1 mainly because it seems like for trezor.io, we need to support subdomains as well. We could of course go full JSON (like request-otr or debounce) and use match patterns, but that's a perf/complexity hit that we should try to avoid if we can. Curious to hear thoughts!

### DIFF
--- a/brave-lists/localhost-permission-allow-list.txt
+++ b/brave-lists/localhost-permission-allow-list.txt
@@ -1,0 +1,6 @@
+! https://community.brave.com/t/bug-brave-shields-break-trezor-bridge-connect-when-accessing-wallet/219544
+trezor.io
+! https://github.com/brave/brave-browser/issues/27346#issuecomment-1435404209
+razroo.com
+! https://community.brave.com/t/intel-driver-support-assistant-doesnt-work-with-shields-on/212468
+intel.com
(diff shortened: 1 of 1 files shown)


antonok-edm (approved):
I'm in favor of a simple text format like this. Personally I'd use `#` as the comment character though, ABP syntax is the only one I know of that uses `!` (not that it matters much, of course :stuck_out_tongue:).

ryanbr (approved):

ryanbr:
Should cover intel.* regional sites @ShivanKaul 

```
intel.de
intel.ie
intel.it
intel.pl
intel.es
intel.com.tr
intel.co.uk
intel.la
intel.com.br
intel.com.au
intel.sg
intel.in
intel.co.id
intel.co.kr
intel.co.jp
intel.vn
intel.com.tw
intel.cn
intel.co.il
```

ShivanKaul:
@pes10k do you mean encoding that a specific website should only be allowed to load localhost resources on a specific port? Currently the permission would be any localhost connection on any port after permission is granted for that website.

pes10k:
ah, i see.  I was still thinking of this in terms of filter lists, and we'd want to be creating rules like: `@@{local}:1234$domain=intel.de`

But if thats not how this new lists' format works, then no worries, no need to complicate things

PR #1164, opened 2023-05-10
Fix https://github.com/brave/brave-browser/issues/19520

ryanbr:
Fixes https://github.com/brave/brave-browser/issues/19520

cc: @arthuredelstein

### DIFF
--- a/brave-lists/https-upgrade-exceptions-list.txt
+++ b/brave-lists/https-upgrade-exceptions-list.txt
@@ -12,6 +12,7 @@ aralego.com
 cri.cn
 frontwing.jp
 chatwork.com
+vger.kernel.org
 fisglobal.com
 fonts.net
 ilive.cn


arthuredelstein (approved):

pes10k:
@arthuredelstein just to check, will the new crawler catch this error case correctly?

arthuredelstein:
In principle it would, yes, because this appears to be a timeout. However this subdomain would need to be on our list of domains.

Given that's not always the case, it does seem we should look into falling back if a site times out after an attempted upgrade. I opened a ticket to work on that: https://github.com/brave/brave-browser/issues/30267

pes10k:
Great! Can you also add this as a possible fail state we can categorize in the study too?

arthuredelstein:
> Great! Can you also add this as a possible fail state we can categorize in the study too?

Will do!

PR #1213, opened 2023-06-14
Remove unique ID from mailchimp "view in browser" pages

fmarier:
Sample URL: https://mailchi.mp/ef7097d62dfa/you-dont-have-a-culture-problem-you-have-a-management-problem?e=a...7

### DIFF
--- a/brave-lists/clean-urls.json
+++ b/brave-lists/clean-urls.json
@@ -243,6 +243,17 @@
     },
     {
         "include": [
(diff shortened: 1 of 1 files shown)


pes10k:
If this is a unique ID, shouldn't we be removing this by default too @fmarier @arthuredelstein ?

fmarier:
https://github.com/brave/brave-browser/issues/31085 for the query filter

PR #1216, opened 2023-06-15
Add matomo params

fmarier:
As documented in https://matomo.org/faq/reports/advanced-how-to-manually-build-campaign-tracking-urls/ and https://help.piwik.pro/support/questions/how-can-i-customize-piwik-pro-campaign-parameters/.

Adguard has some of these: https://github.com/AdguardTeam/AdguardFilters/blob/a10e5b12ce2b7abec0eded17e398729dd0d6bf74/TrackParamFilter/sections/general_url.txt#L199-L210


pes10k:
Some of these, like the `pk_cid` param, seem like we should be removing them by default too, no @fmarier @arthuredelstein ?

fmarier:
https://github.com/brave/brave-browser/issues/31084 for the query filter.

PR #1219, opened 2023-06-16
Add ebay.com and homedepot.com

ryanbr:
Add trackers from ebay.com and homedepot.com


pes10k:
@ryanbr in general, can we keep everything on the first-party-exceptions list tied to specific sites, at least to start? Just want to start off conservative with these rules, and then we can ramp up as we get a sense if we start re-breaking sites

PR #1227, opened 2023-06-20
Add domains linking to localhost

ryanbr:
Add sites from https://gist.github.com/tinogomes/c425aa2a56d289f16a1f4fcb8a65ea65 , Domains linking to 127.0.0.1


pes10k:
Lets hold off on this, since i do not think this will actually solve the uses cases people on EL/EP are asking about. Apologies for jumping the gun on this

ryanbr:
Yeah, close this request. No longer applicable.

Issue #1231, opened 2023-06-22
Add Title and Expires field to Brave lists.


Emi-TheDhamphirInLoveUnderTheFrozenStar:
Brave finally supports the `Expires` property in lists, it would be good if Brave added `Expires:` property and also the `Title` one to the Brave lists, so people adding them the Default lists as custom lists either for testing purposes (like mentioned here https://github.com/brave/brave-browser/issues/29321) or people building Brave from source (https://github.com/brave/brave-browser/issues/26143#issuecomment-1341936225), can have the lists updated more frequently.

But also prettify the Brave lists as Custom so it doesn't look like this:

![image](https://github.com/brave/adblock-lists/assets/122518587/4700462f-720d-4169-a172-4397bb712a5a)

pes10k:
cc @antonok-edm , since i think some of this is either already planned

Emi-TheDhamphirInLoveUnderTheFrozenStar:
@pes10k
If you are talking about the implementation itself, `Expires:` was implemented last week https://github.com/brave/brave-core/pull/18836 and `Title:` in February this year https://github.com/brave/brave-core/pull/17125

So Brave lists only need, for example: 

```
! Title: Brave Unbreak
! Expires: 1 day
```
Also, Brave already have titles for the Brave lists in the [default.json](https://github.com/brave/adblock-resources/blob/master/filter_lists/default.json), so I guess, it's only about copying and pasting them.

Seems like something 'quick n easy', since this will not affect anything about the lists, only the way they look when added as custom and how frequently they automatically update; and then just pushing the update button in the list should reflect these changes immediately.

The thing is mostly how only Brave Team knows how frequent the lists should be updated when added as custom, to avoid problems like high traffic in servers like github that don't belong to Brave, like happened/was reported in October 2022 https://adguard.com/en/blog/easylist-filter-problem-help.html

pes10k:
Understood.  I can only say that, nothing UI related is ever actually "quick n easy" in a code base as big as chromium, and from a position like Brave's (where we need to make sure the modifications we make are as painless-as-possible to apply across chromium versions). As always, PRs welcome ;)

More seriously, like i said, the change you're describing something we're interested in doing.  I dont have a timeline for you though. @antonok-edm may choose to add more information if he has any.

Emi-TheDhamphirInLoveUnderTheFrozenStar:
no no @pes10k the change is in the top part of these Brave lists:
[...]

PR #1234, opened 2023-06-23
Add Title and Expires fields to Brave Default lists

Emi-TheDhamphirInLoveUnderTheFrozenStar:
closes https://github.com/brave/adblock-lists/issues/1231

Adds Title and Expires to the all the the Brave default lists listed in [default.json](https://github.com/brave/adblock-resources/blob/master/filter_lists/default.json) and the [brave-cookie-specific.txt](https://github.com/brave/adblock-lists/blob/master/brave-lists/brave-cookie-specific.txt) list in case that is ever used officially by Brave, so they are readable when added as Custom lists in `brave://settings/shields/filters`.

# Before
image

# After
image

### Note
The only Default list I didn't edit was the **empty** [adblock-lists/brave-lists/brave-unbreak.txt](https://github.com/brave/adblock-lists/blob/master/brave-lists/brave-unbreak.txt)

### DIFF
--- a/brave-lists/brave-cookie-specific.txt
+++ b/brave-lists/brave-cookie-specific.txt
@@ -1,5 +1,5 @@
 ! Title: Brave Cookie-snippet list
-! Description: Used with cookie snippets, set-local-storage-item, set-cookie, set-cookie-reload, set-ession-storage-item
+! Description: Brave-specific additions to Easylist Cookie
 ! Expires: 2 days
 
 ! dealabs mirrors
--- a/brave-lists/brave-cookie-specific.txt
+++ b/brave-lists/brave-cookie-specific.txt
@@ -1,6 +1,7 @@
-! Cookie-snippet list
-! Used with cookie snippets, set-local-storage-item, set-cookie, set-cookie-reload, set-ession-storage-item
-!
+! Title: Brave Cookie-snippet list
+! Description: Used with cookie snippets, set-local-storage-item, set-cookie, set-cookie-reload, set-ession-storage-item
+! Expires: 2 days
+
 ! dealabs mirrors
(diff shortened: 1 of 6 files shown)


pes10k:
this looks great to me, though @ryanbr and @antonok-edm are the experts here.  Thank you for the PR . I see i misunderstood the original ask, so thank you for the time and explaination too

brave-lists/brave-cookie-specific.txt
```diff
@@ -1,6 +1,7 @@
-! Cookie-snippet list
-! Used with cookie snippets, set-local-storage-item, set-cookie, set-cookie-reload, set-ession-storage-item
-!
+! Title: Brave Cookie-snippet list
+! Description: Used with cookie snippets, set-local-storage-item, set-cookie, set-cookie-reload, set-ession-storage-item
```
antonok-edm, line 2: I think better to describe it generally rather than listing specific scriptlet names - e.g. we'll likely move to uBO's `trusted-set-cookie` in the future

something like `Brave-specific additions to Easylist Cookie`

antonok-edm (approved):

antonok-edm:
closes #1231

brave-lists/brave-cookie-specific.txt
Emi-TheDhamphirInLoveUnderTheFrozenStar: Yeah, It crossed my mind that I could have removed the field as well, since only Social had a Description and Brave doesn't do anything with it either (at least for now).
But updated it with your suggestion, so it makes more sense now.
Thanks!.

antonok-edm (approved):

antonok-edm:
Thanks @Emi-TheDhamphirInLoveUnderTheFrozenStar!

PR #1248, opened 2023-07-06
Add youtube scriptlet rule

ShivanKaul:


pes10k (approved):
lgtm

pes10k:
actually, i think all youtube alt domains 301 to youtube.com, and so the scriptlet should only be applied to youtube.com

PR #1283, opened 2023-08-23
Add atinternet params to clean-urls

ryanbr:
Cleanup up any atinternet urls; Based on https://developers.atinternet-solutions.com/as2-tagging-en/javascript-en/campaigns-javascript-en/marketing-campaigns-v2/


fmarier (changes requested):
nit: This would be easier to review if the params were sorted alphabetically.

It looks like this one is missing: `at_send_date`

ryanbr:
Okay, looks good sorted and added at_send_date

fmarier (approved):

PR #1317, opened 2023-09-18
Remove YouTube's new `si` share param

fmarier:
Also add the short domain youtu.be.

Sample URLs:
- https://youtu.be/dQw4w9WgXcQ?si=YjziZuccJeIE-b6z
- https://youtu.be/dQw4w9WgXcQ?feature=shared

Originally reported on brave/brave-browser#33037.

### DIFF
--- a/brave-lists/clean-urls.json
+++ b/brave-lists/clean-urls.json
@@ -156,6 +156,7 @@
     },
     {
         "include": [
+            "*://youtu.be/?*",
             "*://*.youtube.com/watch?*"
         ],
         "exclude": [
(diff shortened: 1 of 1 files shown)


brave-lists/clean-urls.json
```diff
@@ -156,6 +156,7 @@
     },
     {
         "include": [
+            "*://youtu.be/?*",
             "*://*.youtube.com/watch?*"
```
Chiramisu, line 160: I don't believe this will catch `/post/*?`, `/w=*?` (abbreviated param), or even simply `/?` (implied param) links.

fmarier: That's correct. We only clean up the query part of the URL, not the path.

PR #1351, opened 2023-10-13
Add option to disable youtube.com filtering (DO NOT MERGE)

ryanbr:
Will need to be updated, if we need to disable youtube ad filtering.


ryanbr:
No longer needed

PR #1361, opened 2023-10-16
fix https://community.brave.com/t/youtube/509858

Yuki2718:
`trusted-replace-fetch-response` with `$1` seems to cause connection error (you may need to click many videos to reproduce). You can stick to `json-prune` with stack and ignore subsequent uAssets' updates for iOS until we totally switch to  `json-prune` with stack too. Even if YT is watching Brave, I doubt they do on iOS.


ryanbr:
m.youtube.com for iphone?

Yuki2718:
Added

PR #1378, opened 2023-10-27
Fix youtube rules

Yuki2718:
Please test on iPad as I don't have, hope this doesn't cause occasional network error on videos. Another concern is it uses lookahead, do iOS scriptlets support this?


Yuki2718:
Now detected

ryanbr:
cc: @cuba

Yuki2718:
> Another concern is it uses lookahead, do iOS scriptlets support this?

Actually the current rule already uses negative lookahead so I suppose it's okay.

ryanbr:
okay, i guess so

PR #1397, opened 2023-11-06
Fix music.amazon cookies

ryanbr:
Fixes consent message from music.amazon.*;

https://github.com/uBlockOrigin/uAssets/commit/7f35f3140517466f9f7d4cf53b48dbf96aedc984
https://github.com/uBlockOrigin/uAssets/commit/6239f3d6f66792daf31e3a38274c5505dc8a7cb7


Emi-HoloGhostRevisionist88:
Can I ask you why you have to fix this? is it because of iOS? since Desktop and Android support entity for cosmetics and scriptlets, I only can think about iOS needing this fix, so I better ask if I need the information someday in the future.

PR #1402, opened 2023-11-07
Remove unnecessary Instagram parameters related to sharing and embedding

fmarier:
Sample URLs:
- https://www.instagram.com/reel/CyuCMRXPJk6/?utm_source=ig_embed&ig_rid=3d515e33-9a4b-432a-be16-ca0e52a212f5
- https://www.instagram.com/bodyworkbylisa/?ig_rid=a82eb172-092d-4266-89e9-e177b491f610

### DIFF
--- a/brave-lists/clean-urls.json
+++ b/brave-lists/clean-urls.json
@@ -74,6 +74,18 @@
     },
     {
         "include": [
+            "*://www.instagram.com/*"
+        ],
+        "exclude": [
+
+        ],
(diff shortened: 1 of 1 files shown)


fmarier:
`igshid` is already covered by the query string filter (https://github.com/brave/brave-browser/issues/11580) but it might be useful to add it here too in order to clean up same-origin URLs when copy/pasted.

PR #1410, opened 2023-11-15
Update x13_cookies

ryanbr:
Counter the x13 cookies consent. Will remove if uBO takes this up


mapx-:
`click-element` is visually not very nice

```
*##+js(trusted-click-element, button[data-action="choose"], , 500)
*##+js(trusted-click-element, button[data-action="choose"], , 500)
```

mapx-:
`*##+js(trusted-click-element, button[data-action="accept-selected"], , 1000)`

ryanbr:
A bit overeaching?

mapx-:
sure, with all those domains added, otherwise why excessive ? The default seems to be always only about necessary cookies

ryanbr:
Would also assume you accept all the trackers etc, rather than the minimum or necessary. Seems bad for privacy from accept all for everything.

PR #1423, opened 2023-11-24
Address https://github.com/brave/brave-browser/issues/31954

ryanbr:
Fixes login issue on crunchyroll


ryanbr:
Merged, security review not needed. A login fix for crunchyroll due to gpc checks

PR #1455, opened 2023-12-21
Clean ebay links

ryanbr:
Link I found from https://arstechnica.com/cars/2023/12/concorde-engine-sells-on-ebay-may-end-up-as-bits-of-furniture/

`
https://www.ebay.co.uk/itm/116001533010?mkcid=16&mkrid=711-127632-2357-0&ssspo=3X5IDjxbSpS&sssrc=4429486&ssuid=guVqnn7EQqq&var=&widget_ver=artemis&media=COPY
`

Works:
`https://www.ebay.co.uk/itm/116001533010`

### DIFF
--- a/brave-lists/clean-urls.json
+++ b/brave-lists/clean-urls.json
@@ -75,6 +75,23 @@
     },
     {
         "include": [
+            "*://www.ebay.*/itm/*"
+        ],
+        "exclude": [
+
+        ],
+        "params": [
+            "media",
+	    "mkcid",
+            "mkevt",
+            "ssspo",
+            "ssrc",
+            "var",
+            "widget_ver"
+        ]
+    },
+    {
+        "include": [
             "*://www.instagram.com/*"
         ],
         "exclude": [
--- a/brave-lists/clean-urls.json
+++ b/brave-lists/clean-urls.json
@@ -100,7 +100,6 @@
             "ssspo",
             "ssrc",
 	    "ssuid",
-            "var",
             "widget_ver"
         ]
     },
--- a/brave-lists/clean-urls.json
+++ b/brave-lists/clean-urls.json
@@ -75,7 +75,20 @@
     },
     {
         "include": [
-            "*://www.ebay.*/itm/*"
+            "*://*.ebay.com/itm/*",
+            "*://www.ebay.at/itm/*",
+            "*://www.ebay.ca/itm/*",
+            "*://www.ebay.co.uk/itm/*",
+            "*://www.ebay.com.au/itm/*",
+            "*://www.ebay.com.sg/itm/*",
+            "*://www.ebay.de/itm/*",
+            "*://www.ebay.es/itm/*",
+            "*://www.ebay.ie/itm/*",
+            "*://www.ebay.fr/itm/*",
+            "*://www.ebay.it/itm/*",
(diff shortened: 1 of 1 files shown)


brave-lists/clean-urls.json
```diff
@@ -73,6 +73,37 @@
...
+            "*://www.ebay.fr/itm/*",
+            "*://www.ebay.it/itm/*",
+            "*://www.ebay.ch/itm/*",
+            "*://www.ebay.nl/itm/*",
+            "*://www.ebay.pl/itm/*"
+        ],
+        "exclude": [
+
+        ],
+        "params": [
```
fmarier, line 96: Interesting, none of these are on the [AdGuard list](https://github.com/AdguardTeam/AdguardFilters/blob/c1f54ae78eb03858d81a8439999980dd3a57e833/TrackParamFilter/sections/specific.txt#L1351-L1356).

brave-lists/clean-urls.json
```diff
@@ -73,6 +73,37 @@
             "ref_"
         ]
     },
+    {
+        "include": [
+            "*://*.ebay.com/itm/*",
```
fmarier, line 78: Any reason why we shouldn't have the `www` like we do for the other rules?

brave-lists/clean-urls.json
```diff
@@ -73,6 +73,37 @@
...
+
+        ],
+        "params": [
+            "media",
+	    "mkcid",
+            "mkevt",
+            "ssspo",
+            "ssrc",
+	    "ssuid",
+            "var",
```
fmarier, line 103: That's the only one I feel uneasy about. In the example URL we have, it's not set to anything and so it's not clear what it's for or whether or not it's useful. Also, it is **very** generic.

I would leave that one out for now.

fmarier (changes requested):

brave-lists/clean-urls.json
ryanbr: Some regionals; including www. also
```
https://py.ebay.com/
https://pa.ebay.com/
https://sv.ebay.com/
```

brave-lists/clean-urls.json
ryanbr: Can ask Adguard to review this also.

brave-lists/clean-urls.json
fmarier: Ok cool. We should keep the `*` then.

fmarier:
So let's just drop `var` and merge this then.

brave-lists/clean-urls.json
Alex-302: `mkcid`, `mkevt` aff params are ok to block.
Can't find where the rest are used. Could you please describe where can I find them?

fmarier: > Can't find where the rest are used. Could you please describe where can I find them?

We've not found documentation on what these are used for, but you can see an example in the issue description.

fmarier (approved):

brave-lists/clean-urls.json
Alex-302: Ok, added also `ss*` parameters.
Please mention us if users will report about breakage:)

PR #1460, opened 2024-01-02
Few more streaming popups

ryanbr:
A few embedded iframe popups from `https://fmoviesz.to/`

Reported here; https://community.brave.com/t/tabs-for-ads-shopping-sites-automatically-open-when-starting-a-movie-on-fmoviesz-to/


Mikeypdc:
Now it's not just me f moviesz.to but they have a different domain which is fmovies24.to which does the same thing and open another tab with an advertisement in it when you go to click on something on the screen that should just open a movie

PR #1471, opened 2024-01-17
update yt

Yuki2718:
Hopefully should earn time until `uritransform` rule comes to Brave. Also removed duplicate of `youtube.com` in `##+js(brave-fix)`.


ryanbr:
Is this still current @Yuki2718  ?

Yuki2718:
> Is this still current 

Sure, at least until Brave adopt the `uritransform` rule in Quick or we become sure that they no more target cosmetic rules.

PR #1479, opened 2024-01-20
Update yt on iOS

Yuki2718:
Don't know if opening PR makes sense or not as it seems now protected, but anyway, if Brave on iOS now fully supports `trusted-replace-` scriptlet as does uBO, the old json-prune rule which was added only as a workaround when `trusted-replace-` with `$1` caused connection error should be replaced. I suspect this rule may be responsible to occaroinal reports of detection, depending on stack on each user. With fully support I mean https://github.com/gorhill/uBlock/commit/5a24fad8ad8997a40ca2cf3da1bfceb66b326ddb too.


Yuki2718:
cc: @cuba

Yuki2718:
No response? Is YT issues on iOS such a low-priority for Brave?

PR #1485, opened 2024-01-25
Fix Youtube delays on ios

ryanbr:
Fixes perf issues on iOS/youtube, apply older rules


cuba (approved):
Tested this using custom iOS code and works great

PR #1515, opened 2024-02-10
Remove onion V2 address

iam-py-test:
[Onion V2 addresses are no longer accessible](https://support.torproject.org/onionservices/v2-deprecation/), so there is no point in continuing to list them.
I only found one occurrence, but there may be some I missed.
Thank you.


fmarier (approved):

ryanbr:
Addressed in a better PR; https://github.com/brave/adblock-lists/pull/1539

PR #1544, opened 2024-02-22
Mirgate Google signin popup to first-party list

ryanbr:
Possibly fix this; https://community.brave.com/t/is-there-a-way-to-block-this-google-login-thing-that-appears-on-some-websites/532648/2


ryanbr:
Merging, not a sec issue. just wording

PR #1587, opened 2024-03-15
Revert ba.com fixes

ryanbr:
Will need fixes to fingerprinting in shileds. Still causes issues with login with britishairways.com


ryanbr:
Fingerprinting exception will be needed. A new issue report will be logged.

PR #1621, opened 2024-03-28
Add new MailerLite parameters

fmarier:
We added longer parameters for this mailing list service in https://github.com/brave/adblock-lists/pull/1032, but it looks like they're now using a shorter version of them.

Sample URL: https://www.drinkcontainer.beer/csa?ml_sub=24459090&ml_sub_hash=a0o5


ryanbr (approved):
LGTM

PR #1638, opened 2024-04-04
Fixes win7/8 Youtube Crashes

ryanbr:
Interium fix for WIndows 7/8

Removing ' ' quotes on Youtube rules and convert to escape \, to prevent crashing.


ryanbr:
- [x]  Non Premium playback (20 various videos tested)
    - No ads, no delay, instant playback
- [x]  Premium playback (10 various videos tested)
    - No ads, no delay, instant playback

PR #1650, opened 2024-04-09
Merged into Easylist, no longer needed

ryanbr:
Safe to remove? Currently in Easylist, no intention of removing


ShivanKaul:
Yeah, I guess now that it is anchored to Brave Search specifically, it's unlikely to get removed accidentally.

PR #1658, opened 2024-04-14
Fixes cspdailynews.com overlay

ryanbr:
- Fixes overlay on cspdailynews.com
- Also empty space on https://www.supermarketnews.com/r


ryanbr:
no CSP, just part of the domain

PR #1673, opened 2024-04-22
Remove unnecessary params from Narvar tracking URLs

fmarier:
Sample URL: https://ctr.narvar.com/canadiantire/tracking/canadapost?tracking_numbers=29000068&order_number=124058&service=EP&ozip=V1Z%202S3&dzip=V1A4B2&locale=en_CA&origin_country=CA&destination_country=CA

Only `tracking_numbers` and `locale` appear to be necessary for the page to work.

### DIFF
--- a/brave-lists/clean-urls.json
+++ b/brave-lists/clean-urls.json
@@ -379,6 +379,22 @@
     },
     {
         "include": [
+            "*://ctr.narvar.com/*/tracking/*"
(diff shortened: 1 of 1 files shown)


ryanbr:
So none of those params are needed for tracking the package?

ryanbr (approved):

PR #1732, opened 2024-05-12
Move Instacart blocks to ios

ryanbr:
Moved to ios list from being applied to first-party.


brave-lists/brave-ios-specific.txt
```diff
@@ -143,6 +143,11 @@ mp4upload.com##+js(abort-on-property-write, adcashMacros)
 ! apnews.com
 apnews.com##.Page-header-leaderboardAd
 apnews.com##.SovrnAd
+ instacart
```
ShivanKaul, line 146: this will be block all of instacart right? if yes, please revert

PR #1743, opened 2024-05-18
Update twitter rules to include x.com

ryanbr:
- Resync twitter rules in Easylist
- Add x.com and *.x.com too copy clean
- Update twitter-embeds to unbreak

Just keeping twitter.com for the time being, will eventually be pruned to just x.com


brave-lists/clean-urls.json
```diff
@@ -158,6 +158,23 @@
             "t"
         ]
     },
+    {
+        "include": [
+            "*://*.x.com/*",
```
fmarier, line 163: We should add the `x.com` URLs into the above `twitter.com` block to avoid the repetition.

fmarier (changes requested):

brave-lists/clean-urls.json
ryanbr: Sure, merged

brave-lists/clean-urls.json
```diff
@@ -145,10 +145,14 @@
...
         "include": [
-            "*://*.twitter.com/*"
+            "*://*.twitter.com/*",
+            "*://*.x.com/*",
+            "*://x.com/*"
         ],
         "exclude": [
-            "*://*.twitter.com/i/redirect?*"
+            "*://*.twitter.com/i/redirect?*",
+            "*://*.x.com/*",
```
fmarier, line 154: This is going to exclude everything in  the `x.com` domain. Instead, it should be:
```
*://*.x.com/i/redirect?*
```

brave-lists/clean-urls.json
```diff
@@ -145,10 +145,14 @@
     },
     {
         "include": [
-            "*://*.twitter.com/*"
+            "*://*.twitter.com/*",
+            "*://*.x.com/*",
```
fmarier, line 149: From what I can see in https://chromium.googlesource.com/chromium/src/+/main/extensions/common/url_pattern.cc, `x.com` is already included in `*.x.com` and so we don't need an explicit `x.com`.

fmarier (changes requested):

fmarier (approved):
`brave-lists/clean-urls.json` changes look good

PR #1753, opened 2024-05-22
Test filter for youtube nags

ryanbr:
Fixes: Also is shared with other youtube nags using different wording.

image (16)
image (17)


ryanbr:
Cover 192px, and 191px

```
tp-yt-paper-dialog style-target="host" role="dialog" tabindex="-1" class="style-scope ytd-popup-container" modern="" style="outline: none; position: fixed; inset: auto auto 0px 0px; margin: 16px; z-index: 2202; box-sizing: border-box; max-height: 192px; max-width: 2484px;"
```

PR #1767, opened 2024-05-30
Exempt mitmproxy installation URL

fmarier:
See https://docs.mitmproxy.org/stable/overview-getting-started/


arthuredelstein (approved):
thanks!

PR #1812, opened 2024-06-16
Address YT rules without extended args

ryanbr:
Add excepton rule for the` , sedCount, 1)` in both ios and unbreak
and add the same rule without these (Addressing issue on https://github.com/brave/adblock-rust/issues/376)


ryanbr:
do not merge until we know for sure

PR #1836, opened 2024-06-27
Fixes visortmo.com popups and cosmetics

ryanbr:
Hide the ads on visortmo.com and prevent popups.


ryanbr:
Address https://github.com/brave-experiments/web-compat-reports/issues/537

PR #1858, opened 2024-07-03
Debounce Selzy link tracker

fmarier:
This is the link tracker used by this email marketing platform: https://selzy.com/

Sample URL: https://link.emlmind.com/en/mail_link_tracker?hash=66c5ibqsukgp7ikdzwu4ujmw7y7boyubuqqmaq86u5sjqdohiizkiackas9r171wq1go8tfs5eongbftyy5rh47j45mgawwfzgn66xho&url=aHR0cHM6Ly93d3cubWVsb3R0b2dyb3VwLmNvbS9zY2hlZHVsZQ~~&uid=Njg1OTgwOA~~&ucs=88ff10da1ab90c2435eec8aa3fb9fb04

### DIFF
--- a/brave-lists/debounce.json
+++ b/brave-lists/debounce.json
@@ -351,6 +351,15 @@
   },
   {
     "include": [
+      "*://link.emlmind.com/en/mail_link_tracker?*"
+    ],
+    "exclude": [
+    ],
+    "action": "base64,redirect",
+    "param": "url"
+  },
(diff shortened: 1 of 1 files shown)


fmarier:
This doesn't actually work because they're using `~~` instead of `==` in the base64 encoding. We'd have to do a string replacement before the base64 decode.

PR #1912, opened 2024-07-24
Ensure Brave component updater isn't accidentally blocked by a filter rule

ryanbr:
Preeemptive exception, in the case of a list (any list) accidently blocking the updator.


ShivanKaul:
Shouldn't these live in https://github.com/brave/adblock-lists/blob/master/brave-lists/brave-specific.txt

ryanbr:
We've kept these in unbreak for some reason, but happy to move them over

ShivanKaul:
okay let's just move them all over

PR #1939, opened 2024-08-03
Fix youtube.com ads in Brave release

ryanbr:
Will fix most of the YT ads being reported.


brave-unbreak.txt
```diff
@@ -5,6 +5,13 @@
...
+
+!
+! Set trusted types scriptlet on Youtube
+! www.youtube.com##+js(set-trusted-types)
```
ShivanKaul, line 11: Why is this commented out?

ryanbr: - It wasn't doing anything we needed to change it too `www.youtube.com##+js(trusted_types)`
- Also Gorhill up'd changes making it https://github.com/gorhill/uBlock/commit/4f0d1301ab48e62afb3424609834c6353a0a5eee making the sciptlet not needed. Which we have sync'd. 
- Still working out what is actually broken, since the rules worked fine previously. Both trusted-rpnt sciptlets just broke

PR #1944, opened 2024-08-06
youtube quality android

tapanmodh:
Related to issue: https://github.com/brave/brave-browser/issues/32842
Brave-core PR: https://github.com/brave/brave-core/pull/24164
Adblock-resource PR: https://github.com/brave/adblock-resources/pull/196

### DIFF
--- a/brave-lists/brave-specific.txt
+++ b/brave-lists/brave-specific.txt
@@ -45,6 +45,9 @@ youtube.*##+js(brave-youtube-theater-fix)
 ! YouTube navigation fix scriptlet rule
 youtube.*##+js(brave-youtube-navigation-fix)
 
+! YouTube video quality
+youtube.*##+js(brave-youtube-quality)
+
 ! Used for Brave QA tests
 ?335962573013224749$image,xhr,domain=dev-pages.brave.software|dev-pages.bravesoftware.com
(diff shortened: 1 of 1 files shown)


ShivanKaul:
This rule will apply immediately across all Brave versions, whereas https://github.com/brave/brave-core/pull/24164 will only go into Nightly. Is that the intention?

tapanmodh:
> This rule will apply immediately across all Brave versions, whereas [brave/brave-core#24164](https://github.com/brave/brave-core/pull/24164) will only go into Nightly. Is that the intention?

Is it? We need it with [brave/brave-core#24164](https://github.com/brave/brave-core/pull/24164) as this PR handles settings for that. Can we apply this rule to only Nightly first?

ryanbr:
Maybe apply to specifically `m.youtube.com` to limit the scope to mobile only

ShivanKaul:
> Can we apply this rule to only Nightly first?

There's no way to apply a filter rule to a particular channel. You could however create a new list and enable it on Nightly via Griffin, see: https://github.com/brave/brave-core/pull/22858/files
This issue captures the series of steps for introducing a new list, adding it to the filter catalog, and selectively turning it on in a specific channel: https://github.com/brave/brave-browser/issues/36244

PR #1946, opened 2024-08-07
Add new YT rule for IOS

ryanbr:
Updated YT anti-ad rule


ryanbr:
Adjusted PR from experimental to IOS.

https://github.com/uBlockOrigin/uAssets/commit/ab812e099 was released into uBO, no need for experimental.

PR #1971, opened 2024-08-19
Attempted fix for pornhub.com

ryanbr:
Attempted fix for `pornhub.com`/`es.pornhub.com`


ShivanKaul:
But this won't fix https://github.com/brave-experiments/web-compat-reports/issues/683 right? Those are on brave-core

ryanbr:
Sorry for the late response, just noticed the reply. Seems it did fix it. Was related to some 1p pornhub scripts. @ShivanKaul

PR #1976, opened 2024-08-21
Update Ad-shield on ios

ryanbr:


Yuki2718:
Does `*$doc,csp=script-src-attr 'none',to=badmouth1.com|eurointegration.com.ua|flatpanelshd.com|hoyme.jp|issuya.com|iusm.co.kr|jin115.com|lamire.jp|picrew.me|sportsrec.com` work on iOS while not on other platforms?

PR #1983, opened 2024-08-23
Sync AS

Yuki2718:
https://github.com/uBlockOrigin/uAssets/commit/8f03f241e509e93122befc8b95f95a5b1b6d1339
We partly moved to redirect + scriptlet approach to minimize false positive.


ryanbr:
No longer needed atm

PR #2022, opened 2024-09-04
fix AS for iOS

Yuki2718:
All thanks to AdGuard, (hopefully) will fix https://community.brave.com/t/topic/562179 and many more, but I have no way to check on real device. Currently Ad Shiled ads are mostly hidden by generic cosmetic filters, but they are not perfect and leave placeholder and let their tracking run.


Yuki2718:
Reportedly no longer works.

Yuki2718:
AG says it works.

PR #2057, opened 2024-09-21
Revert YT change, use previous working rule

ryanbr:
Due to the feedback of YT ads being reported, we'll revert back to previous rule

https://community.brave.com/t/brave-stopt-blocking-youtube-adds/569660/26?u=fanboynz


tomocrafter:
Thank you for update.
It seems removed ads but now End Screen appear way faster than it should. It happens without this new script but does not happen if Brave Shields is turned off. No extensions is enabled.

PR #2121, opened 2024-10-26
Fix YT rule

ryanbr:
We don't support the first quote in the YT rpnt scriptlet.


antonok-edm (approved):

Yuki2718:
Addressed in upstream, was a bad timing https://github.com/uBlockOrigin/uAssets/commit/7731b1a6e60cdfb7fa32b5b79f86ba556b218447

PR #2151, opened 2024-11-15
Address cinego.tv popups

ryanbr:
Address https://github.com/brave-experiments/web-compat-reports/issues/981


ShivanKaul:
Just curious, why are we addressing popups in this list?

PR #2163, opened 2024-11-18
Remove Threads' XMT Tracker

jonathansampson:
Remove the xmt-tracker from Threads' "copy link" button.

Resolves https://github.com/brave/brave-browser/issues/42345

### DIFF
--- a/brave-lists/clean-urls.json
+++ b/brave-lists/clean-urls.json
@@ -843,6 +843,17 @@
     },
     {
         "include": [
+            "*://*.threads.net/*"
+        ],
+        "exclude": [
+
+        ],
+        "params": [
+            "xmt"
+        ]
+    },
+    {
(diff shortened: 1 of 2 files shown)


brave-lists/clean-urls.json
```diff
@@ -852,5 +852,16 @@
             "brave_testing1",
             "brave_testing2"
         ]
+    },
```
fmarier, line 855: Would you mind moving this up so that we keep our test entries last in the file?

fmarier (changes requested):
One small change, but otherwise looks good.

brave-lists/clean-urls.json
jonathansampson: Done!

fmarier (approved):

PR #2171, opened 2024-11-21
Debounce search.app

Emi-Emi-Emi:
Sample URL:  https://search.app/?link=https%3A%2F%2Fwww.ghacks.net%2F&utm_campaign=aga&utm_source=agsadl2%2Csh%2Fx%2Fgs%2Fm2%2F4

### DIFF
--- a/brave-lists/debounce.json
+++ b/brave-lists/debounce.json
@@ -170,7 +170,8 @@
   },
   {
     "include": [
-      "*://*.getprice.com.au/prodhits.aspx?*"
+      "*://*.getprice.com.au/prodhits.aspx?*",
+      "*://search.app/?*"
     ],
     "exclude": [
     ],
--- a/brave-lists/debounce.json
+++ b/brave-lists/debounce.json
@@ -170,8 +170,8 @@
   },
   {
     "include": [
-      "*://*.getprice.com.au/prodhits.aspx?*",
-      "*://search.app/?*"
+      "*://search.app/?*",
+      "*://*.getprice.com.au/prodhits.aspx?*"
     ],
     "exclude": [
     ],


Emi-Emi-Emi:
This was reported by this [ghacks article](https://www.ghacks.net/2024/11/08/are-links-beginning-with-search-app-safe/) where official Google app for Android will generate share links as _search.app_ url.

fmarier:
@Emi-Emi-Emi Have you been able to reproduce it? I've tried setting Chrome as the Browser app on my Android phone and sharing links in it doesn't use `search.app` for me.

Your PR looks good, but I need to test sharing a URL which has query string parameters already to confirm that it escapes the parameters and that they are included in the `link` attribute.

Emi-Emi-Emi:
People said it is from the Google app, not Chrome, but I couldn't reproduce it in the Android phone I tested, I even downgraded the Google app to `15.44.27.28.arm64` based on [Bleepingcomputer Article](https://www.bleepingcomputer.com/news/security/googles-mysterious-searchapp-links-leave-android-users-concerned/), so I don't know if it is a region or Android OS thing, but I saw the ghacks article and then some people in reddit talking about it.

But the best way to find links and see how many people are sharing these links was X though, just searching for `search.app` (https://x.com/search?q=search.app&src=typed_query&f=live) will show hundreds of links being shared daily by people who do it unknowingly (I guess) and that means anyone clicking on it will be sent to Firebase/Google domain before being redirected to the actual site.

It seems there are two types of shareapp links though, one like `https://search.app/EFEaa9zy4Y3XyoAC9`

but then more of those links are the ones that can be _debounced_ like https://x.com/yukikazemaru/status/1859734241430601913 with:
- https://search.app/?link=https%3A%2F%2Fwww.digital.archives.go.jp%2FDAS%2Fmeta%2FDetail_F0000000000000104820&utm_campaign=aga&utm_source=agsadl2%2Csh%2Fx%2Fgs%2Fm2%2F4

fmarier:
I found this link on Twitter: `https://search.app/?link=https%3A%2F%2Fallthingslawandorder.blogspot.com%2F2018%2F01%2Flaw-order-svu-info-wars-recap-review.html%3Fm%3D1&utm_campaign=aga&utm_source=agsadl2%2Csh%2Fx%2Fgs%2Fm2%2F4`
which goes to `https://allthingslawandorder.blogspot.com/2018/01/law-order-svu-info-wars-recap-review.html?m=1` and so it does url-encode query string params into the `link` attribute.

brave-lists/debounce.json
```diff
@@ -170,7 +170,8 @@
   },
   {
     "include": [
-      "*://*.getprice.com.au/prodhits.aspx?*"
+      "*://*.getprice.com.au/prodhits.aspx?*",
+      "*://search.app/?*"
```
fmarier, line 174: Can you please insert it at the top of the list, just before the other entry so that we don't have to touch the existing entry (and add noise to the `git blame`)?

fmarier (commented):
One small thing and that it's good to go.

brave-lists/debounce.json
Emi-Emi-Emi: Okay! Done.

fmarier (approved):
Thanks!

PR #2229, opened 2024-12-22
Update streaming sites

ryanbr:
Re-sync reliabletv.me fix
Update regex from uBO
Remove vipbox/fbstreams to avoid possible issues


brave-unbreak.txt
```diff
@@ -17,7 +17,8 @@
 .shop/cuid/
 
 ! workaround https://github.com/brave/brave-browser/issues/42496
-reliabletv.me##+js(rpnt, script, /function isBrave[\s\S]*(?=const doAuthCheck)/)
+! reliabletv.me##+js(rpnt, script, /function isBrave[\s\S]*(?=const doAuthCheck)/)
+reliabletv.me##+js(rpnt, script, /const doRedirect = 1;[\s\S]*(?=const doAuthCheck)/, const doRedirect = 1;)
```
Emi-Emi-Emi, line 21: @ryanbr they changed the script again, first they reverted to use the simpler `isBrave`, but as `isBraveBrowser` and but now they are also moving the script around so the lookahead (?=) will not work.

This is stupid. BTW, maybe you should push the support of regex in the domain and cosmetics, one way to 'mask' these type of things is by using regex even if slower, so the rule will not be reliabletvme which can be searchable.

You can do this:

`reliabletv.me##+js(rpnt, script, /const [a-z]{2}B.+((\n).*){8\,11}(f\w{4}K\w[l]{2}.+\,\s[e-u]{4}\);((\n).*){1}})/)`

So at least they will not spot it as easy.

or `/const isBraveBrowser[\S\s]*(?=const secTokenUrl)/` but it would be easy to see for anyone.

This is why it sucks adblocking lists are too public, I doubt they are seeing list.txt from Brave, they might just see the same line in brave-unbreak and just check what it does now and modify it, so I guess even a complex regex will not matter.

This is why a domain with regex would be ideal, at least as a way to hide it a little, because these people are desperate to really see what brave is doing.
I wish these pages would stop using relabletv, but yeah, I am sure they don't care.

I forgot I can send comments in PRs, so I think it is better than commenting on the issue that they probably also check, it's obvious they knew it was updated. But you adding the link to the issue also didn't help.

I just wish anti-Brave mob would stop. But seems like you have to talk to the other people to really deal with that Sec-UA BS detection, Brave has to give up with its own useragent, sec-ua and everything that can detect it as not Chrome, there is no way to defeat this unless Brave adds a natie way of modifying the headers, and all brave talked (as linked in the issue) was allowing to remove things like cookies like uBlock does, extensions are bad unless mobile supported them.

Emi-Emi-Emi: I also took a look more about the code and this seems like a better approach to remove the part that will redirect Brave users to the message html, if there is nothing to redirect to, then nothing happens.

`(rpnt, script, window.location = '/techissue.html';)`

and this is how I hid it on my side even if they can't see my fix.

`reliabletv.me##+js(rpnt, script, /w[a-z]{5}\.loc[a-z]{5} = '\/t.+;/)`

or a more reduced way because there is only one place to match this  `##+js(rpnt, script, /.+'\/t.+'/)`

PR #2240, opened 2024-12-30
Fixes detection

asihd0as8hd8shad:


Emi-Emi-Emi (commented):
`             `

Emi-Emi-Emi:
the last fix was detected within 12 hours, last week. I am sure they are watching either the list or the issue with my comment about it so they just did some changes and done making the regex useless.

I went ahead and have been using this fix for a week, and it will be better since it deals with the redirection part to the respective html, forcing them to do something else. (in other words, it is a properly made fix, unlike before).

Emi-Emi-Emi:
Fix for the anti-Brave redirection that happens in many sport streaming sites.

Emi-Emi-Emi:
BTW, for some reason VSCode glitched and didn't add the description, and I added a comment but it added it as a 'review comment' which cannot be deleted or at least by me, so if you can remove it, please do so, since it wasn't meant to be that way.

Emi-Emi-Emi:
Also, this fix is important, and since the Sec-CH-UA check is still done in JS, it is good there can be a fix (especially for mobile users who cannot install extensions), unlike others that are using the check on the server.

The fix takes care of removing the HTML part in the redirect, that means, unlike previous rules, this should be better.

It is clear they are reading my previous comments from the issue, so I couldn't comment there to provide the new fix, they first scrambled the isBrave because I mentioned that, and they probably saw that it was useless so they just switched lines around so the regex wouldn't work, in this case it will just find any .html'; line and remove it, which should be better since they would have to do something else to create the redirection.

It is still a workaround because Brave has to decide about Sec-CH-UA since Brave is being targeted with it, but being based on JS is good because Adblocker can still deal with it.

Please review and merge if possible.

I didn't care about this fix but since they are really watching to block Brave, then it became a little more personal. I am also using my own fixes for this so others can use these sites in Brave, so I am also risking my fix by providing it to others, and hopefully they will not go further with the block.

Thank you.

Emi-Emi-Emi:
So what is the point of Brave adblock lists being in Github and letting people have access to push PRs, if nobody is going to even take a look at them and say yes or no, and close or merge?

This is actually will fix it and help people until Brave finally gets rid of Brave in the `Sec-CH-UA`, which is the only way they can block Brave.
So, why should I write a comment in an issue that is obviously the guys doing the anti-Brave are reading, just to get the fix kill in 12 hours like happened the last time I did that? If not just restrict the access to people. 36 issues and 14 PRs, many will not get closed or touched, this is the problem with Brave team... how hard is to just clean things and make sure good PRs get merged?

Not all people working for Brave don't care about issues and PRs, but some apparently don't really care if you still have issues that should have been closed as "resolved" or "duplicates" or "won't fix" or "depends on this other feature [LINK]".

Either merge or close this, or do the same I did, find the script, do the rpnt, until next time when they get to block Brave again, until they start obfuscating the code and making it harder and then do the server side block like weirdos at lobsters do.

Emi-Emi-Emi:
And don't get me wrong, this is not "oh do what I say" it's just that it is a waste of time for me to even try to do anything, and seeing how there is a PR from november 12th, that never got closed as invalid because Brave doesn't even support denyallow, it's just clear how truthful my message is.

Do I have to tag 800 people just to get something merged that takes 5 seconds to test?

Emi-Emi-Emi:
just go to https://www.vipbox.lc/golf/nbc-golf-channel-1-live

test the rule, it should work. 

remove the rule, it should say some dumb message about having to use other browsers like if you couldn't install adblockers there and block the same ads.

PR #2301, opened 2025-02-07
Add exceptions for statcounter

pes10k:
Add exceptions for statcounter after auditing that those resources can be loaded safely and with allowing Brave users to be tracked cross-site or cross-profile (i.e., they do not try to circumvent brave's tracking protections)

### DIFF
--- a/brave-lists/brave-specific.txt
+++ b/brave-lists/brave-specific.txt
@@ -10,6 +10,15 @@
 ||[::1]^$third-party,domain=~localhost|~127.0.0.1|~[::ffff:7f00:1]
 ||[::ffff:7f00:1]^$third-party,domain=~localhost|~127.0.0.1|~[::1]
 
+! Allow page views in Brave to be counted by statcounter.com. The scripts that match the first
+! two rules below have been audited to confirm they don't try to circumvent Brave's cross-site
+! or cross-profile tracking protections (e.g., they don't try to fingerprint users, or detect
+! private mode, etc.). Brave's other protections (storage partitioning, empheral site lenght
+! storage, etc.) prevent statcounter code from tracking or profiling users. 
+@@||secure.statcounter.com/counter$script
+@@||www.statcounter.com/counter$script
+@@||c.statcounter.com^$image
+
 ! brave.com specfic filters
 @@||ads.brave.software^$first-party
 @@||ads-admin.bravesoftware.com^$first-party


brave-lists/brave-specific.txt
```diff
@@ -10,6 +10,15 @@
 ||[::1]^$third-party,domain=~localhost|~127.0.0.1|~[::ffff:7f00:1]
 ||[::ffff:7f00:1]^$third-party,domain=~localhost|~127.0.0.1|~[::1]
 
+! Allow page views in Brave to be counted by statcounter.com. The scripts that match the first
+! two rules below have been audited to confirm they don't try to circumvent Brave's cross-site
+! or cross-profile tracking protections (e.g., they don't try to fingerprint users, or detect
+! private mode, etc.). Brave's other protections (storage partitioning, empheral site lenght
```
ShivanKaul, line 16:
```suggestion
! private mode, etc.). Brave's other protections (storage partitioning, ephemeral
```

ShivanKaul (approved):
We'd want to follow this up with:
1.  A resource/script replacement to lock down the privacy risk even more (our privacy review of their script is a point-in-time check)
2. Don't add these exceptions for users in Aggressive ad & tracker mode. We don't currently have the ability to only apply a rule or a filter list in one mode but not the other.

In the meantime, users can `badfilter`  these rules (i.e. make sure that the exception doesn't apply) by adding the following custom rules in the **Settings > Shields > Content filtering > Custom filters** (might have to enable developer mode):
```
@@||secure.statcounter.com/counter$script,badfilter
@@||www.statcounter.com/counter$script,badfilter
@@||c.statcounter.com^$image,xhr,badfilter
```

ryanbr:
Guessing can't we limit it to a specific domain?
Would still require statcounter to add support for Brave

ryanbr (commented):
Not sure, what I saw from statcounter (sample, https://unsri.ac.id/ )
```
@@||secure.statcounter.com/counter/counter.js$script
@@||c.statcounter.com/t.php$xhr
```

pes10k:
just a note @ShivanKaul that I made a small change to also exempt ajax calls to c.statcounter.com, since statcounter collects pings using XMLHTTPRequest calls from some initial scripts

pes10k:
> Not sure, what I saw from statcounter (sample, https://unsri.ac.id/ )
> 
> ```
> @@||secure.statcounter.com/counter/counter.js$script
> @@||c.statcounter.com/t.php$xhr
> ```

These will be too narrow.  The former one will miss the `counter_xhtml.js` variant for the initial code, and the later one will miss the `<noscript>` image for the ping call.

It'd be great to be able to strip query parameters from these calls though. If there is a way to do that @antonok-edm @ryanbr please let me know and i can futher update the PR

antonok-edm:
It'd be great to be able to strip query parameters from these calls though. If there is a way to do that

`$removeparam`: https://github.com/gorhill/ublock/wiki/Static-filter-syntax#removeparam

antonok-edm:
> In the meantime, users can `badfilter` these rules (i.e. make sure that the exception doesn't apply) by adding the following custom rules in the **Settings > Shields > Content filtering > Custom filters** (might have to enable developer mode):
> 
> ```
> @@||secure.statcounter.com/counter$script,badfilter
> @@||www.statcounter.com/counter$script,badfilter
> @@||c.statcounter.com^$image,xhr,badfilter
> ```

`$badfilter` only works within the same engine; the custom filters get added to the "no first-party protections" engine but `brave-specific.txt` filters only get added to the other engine with first-party protections.

The proper way to override the exception rules would be by adding the custom rules, but unexcepted and with an `$important` modifier:

```
||secure.statcounter.com/counter$script,important
||www.statcounter.com/counter$script,important
||c.statcounter.com^$image,xhr,important
```

pes10k:
> > It'd be great to be able to strip query parameters from these calls though. If there is a way to do that
> 
> `$removeparam`: https://github.com/gorhill/ublock/wiki/Static-filter-syntax#removeparam

This is currently only applied in aggressive mode though right? And since we will just (at some point, even if not v1) outright block these requests in aggressive mode, that might be a pickle

that said, i think this would be good to add in the meantime. I’ll update the PR tomorrow

pes10k:
On re-review, the lines of code i thought were pushing query parameters along don't copy-over the original values, just that there are values, so i dont think theres a need for `removeparam` after all, my mistake:

for example:

```
                for (var i = ["utm_source=bing", "?gclid=", "&gclid=", "?msclkid=", "&msclkid=", "utm_medium=cpc", "utm_medium=paid", "utm_medium=ppc"], a = 0; a < i.length; a++)
                    if (g.location.search.includes(i[a])) return "p";
                for (var f = ["utm_medium=email"], a = 0; a < f.length; a++)
                    if (g.location.search.includes(f[a])) return "e";
                if (!I(e) && n == r) return "internal";
```

PR #2333, opened 2025-02-24
Reported perf issues on https://steelseries.com/

ryanbr:
User reported perf issue, links show up slowly when clicked. Works fine in Aggressive mode.

Site is using a 1p google tracking script.

`https://steelseries.com/gaming-mousepads/qck-heavy-xxl-destiny-2-final-shape-edition` Clicking through various menu items, its slow in Brave


ShivanKaul:
In the future we shouldn't add generic rules to this list without discussion @ryanbr

PR #2400, opened 2025-04-01
Add Amazon Ireland and Amazon South Africa

fmarier:
Amazon South Africa opened in May 2024 and Amazon Ireland in March 2025.


fmarier:
Sadly no amazon.co.nz yet :(

ryanbr (approved):

PR #2410, opened 2025-04-06
Update Ad-Shield

Yuki2718:
Quick-fixes blocks some 1pt requests of `html-load.com` with `*$~script,~frame,~xhr,domain=07c225f3.online|content-loader.com|css-load.com|html-load.com|img-load.com` and it doesn't depend on any advanced features like scriptlet. This PR allows to keep them blocked. Also if you need the `html-load.com` exception, you'll also need an exception for their legacy domain `07c225f3.online` used e.g. on `sportalkorea.com`.


Yuki2718:
`07c225f3.online` is probably no longer in use, I'm fine to remove if requested.

PR #2433, opened 2025-04-21
added filter list for yt playables

rakicjovan:
Added a filter list for blocking the "Youtube Playables" section.


ryanbr:
Already in Brave now, safe to close. brave://settings/shields/filters

PR #2454, opened 2025-05-08
Update generic block

ryanbr:
Add a couple of generic adserver blocks


ryanbr:
Okay, fixed my verified.

PR #2508, opened 2025-06-12
Add lucid.app to localhost permission allow-list

drmercer-lucid:
Lucidchart and Lucidspark have a "Load local extension" feature that developers can use when working with the Extension SDK. This requires making a connection to `http://localhost:9900`. See [the Extension SDK documentation](https://developer.lucid.co/docs/extension-api-getting-started#debug-your-editor-extension) for more details.

Figma, a similar SaaS product, is already on the list.


ryanbr:
Localhost requests will handled with a permission popups now, safe to close now

PR #2509, opened 2025-06-13
Disable google.com consent rules for the moment

ryanbr:


ryanbr:
Possibly related too https://github.com/brave/brave-browser/issues/46683

PR #2520, opened 2025-06-18
Remove `media devices` from list of exceptions

ShivanKaul:
Not supported (should be `media-devices`). If it's not supported, seems fine to remove.


ryanbr (commented):
If its a typo, should it be changed media-devices rather than removed?

ShivanKaul:
> If its a typo, should it be changed media-devices rather than removed?

But is it even required? I would guess that the website is fixed without the exception? Can you double-check?

PR #2522, opened 2025-06-19
Validate that exceptions in `webcompat-exceptions.json` are supported

ShivanKaul:
Adds a new GitHub Action task to validate that all listed `exceptions` in `webcompat-exceptions.json` are supported (listed in https://github.com/brave/brave-core/blob/master/components/webcompat/content/browser/webcompat_exceptions_service.cc). We can improve this script to add additional validations later. Update: also validates that `jq` runs correctly.

I tested the Action locally with `act`.

This will fail until we resolve https://github.com/brave/adblock-lists/pull/2520 (done)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,8 +25,11 @@ jobs:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
-      - name: Install yajl-tools
-        run: sudo apt-get install -qq yajl-tools
+      - name: Install list-validation tools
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -25,8 +25,11 @@ jobs:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
-      - name: Install yajl-tools
-        run: sudo apt-get install -qq yajl-tools
+      - name: Install list-validation tools
```
ShivanKaul, line 28: Needed to make sure `jq` was available.

antonok-edm (approved):

PR #2523, opened 2025-06-19
Use hyphens for multi-word exception strings

arthuredelstein:
Resolves https://github.com/brave/brave-browser/issues/46652


ShivanKaul:
If the exception was never working in the first place, then do we even need this rule?

ryanbr:
I'll check next week, just away from the PC

bsclifton:
cc: @ryanbr @ShivanKaul is this still relevant?

PR #2557, opened 2025-07-17
Update yt-shorts.txt

mindplay-dk:
update blocking rules found here:

https://www.reddit.com/r/Adblock/comments/ts7mia/youtube_shorts_custom_filter_blocks_all_shorts/

(the current ones do not block anything on the main page)

### DIFF
--- a/brave-lists/yt-shorts.txt
+++ b/brave-lists/yt-shorts.txt
@@ -1,7 +1,7 @@
 ! Youtube Shorts
 ! Main page
-youtube.com##ytd-grid-video-renderer:has(#thumbnail[href*=shorts])
-youtube.com##ytd-grid-video-renderer:has(#video-title:has-text(/#shorts/i))
+www.youtube.com##ytd-rich-section-renderer:has(#icon)
+www.youtube.com##.ytGridShelfViewModelHost:has(.ytIconWrapperHost)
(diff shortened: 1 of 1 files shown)


mindplay-dk:
Hmm, this worked great for a few days, and now it doesn't. 🤨

How do you develop and test ad filter rules?

I was surprised to find there's a whole graphical UI that generates ad-filters *on the mobile version of Brave only?*

How putting something in dev tools, or an extension, something, anything so we can develop and test ad rules?

What do people do now? Just keep editing and reloading? That doesn't work - especially on something like YouTube where everything is different every time you reload the page... 😌

mindplay-dk:
how about adding "Copy adblock filter" here?

image

I know you have the functionality internally somewhere, because the feature is exposed in the mobile version. (again, no idea why the mobile version would have this feature before the desktop version - developers don't create these filters from their mobile phones, do they?)

mindplay-dk:
Hmm, no, this doesn't work anymore either.

The one that works for me is `www.youtube.com##ytd-rich-section-renderer:has(#icon)` - which seems dangerously broad, but the shorts shelf seems to be the only one that has the icon for now.

I can't find any other rule that works.

ryanbr:
YT Shorts sorted now, via shields.

PR #2572, opened 2025-07-31
Fix Chrome popup

ryanbr:
Fix "Choose Chrome" popup on google.com

chrome-popup


ryanbr:
If this looks good @ShivanKaul @antonok-edm

Rosenkreutzyami:
gh pr checkout 2572

PR #2588, opened 2025-08-13
chore: Update community domain name

jwadolowski:
For https://github.com/brave/devops/issues/13832

The redirect is included, so all the links will still work. Merge at your convenience.


ShivanKaul (approved):
Thanks for this

PR #2611, opened 2025-09-07
Add brave-auto-pip scriptlet rules for YouTube and Brave Talk

thypon:
Applies the brave-auto-pip scriptlet to automatically enable picture-in-picture functionality on:
- All YouTube domains (youtube.*)

This provides seamless automatic picture-in-picture experience for video content on these platforms.

Resource PR: https://github.com/brave/adblock-resources/pull/273

### DIFF
--- a/brave-lists/experimental.txt
+++ b/brave-lists/experimental.txt
@@ -24,6 +24,9 @@ www.google.com,www.google.ca,www.google.com.au,www.google.co.nz,www.google.co.uk
 ! Test rule to prevent open-in-app on mobile
 m.youtube.com##+js(rc, paused-mode, , stay)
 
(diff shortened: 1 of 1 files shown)


ShivanKaul:
What happens on mobile? Is this only for desktop?
Given that this is YouTube we need to be very cautious; there's a lot of rules and scripts uBO and we inject so there's a high risk of a bad interaction. Can you include screen recordings for with and without the scriptlet? You can use custom scriptlets to test. We definitely need @ryanbr  to weigh in here.

thypon:
this is all try caught. It should just fail if no support is available.

ryanbr (approved):

PR #2636, opened 2025-10-07
Fixes ad popups on https://dlhd.dad/watch.php?id=39

ryanbr:
- Fixes div css overlay (imported from uBO)
- Add HEAD ad popup


ShivanKaul:
The upstream rule that fixes this and that we don't currently have support for is `/^https?:\/\/[a-z]{8,15}\.net\/$/$xhr,3p,method=head`.

Support for that rule in adblock-rust is being tracked in https://github.com/brave/adblock-rust/issues/391.

brave-unbreak.txt
```diff
@@ -1418,11 +1418,15 @@ streamwish.to##+js(abort-on-property-read, __Y)
 ! method:HEAD regex
 /^https?:\/\/[a-z]{8,15}\.(com|xyz)\/$/##+js(no-fetch-if, method:HEAD)
 ! embasic.pro (HEAD ADS work around)
```
ShivanKaul, line 1420:
```suggestion
! embasic.pro (HEAD ADS work around)
! See https://github.com/brave/adblock-rust/issues/391
```

ShivanKaul:
Looks like to support this rule we also need: https://github.com/brave/adblock-rust/issues/257

PR #2646, opened 2025-10-18
Add new domains to brave-checks.txt

Zamanhuseyinli:
Spam gambling site ads and special domains designed for advertising purposes have been added.


ryanbr:
Brave-checks isn't what this is used is for. Where do these domains orginate from?

Zamanhuseyinli:
These domain names come from more than one site I've come across, man. And doesn't Brave offer ad blocker support to prevent spammy ads?

PR #2648, opened 2025-10-19
Address canvas issue on meet.google.com

ryanbr:
Address https://github.com/brave/brave-browser/issues/32836


ShivanKaul (approved):
Let's please check with users on https://github.com/brave/brave-browser/issues/32836 if this actually fixed the problem for them.

PR #2651, opened 2025-10-21
Remove old Google consent exceptions

ryanbr:
Remove dead google consent exceptions
Add a longer click delay for google consent, from 500->1800ms.


ShivanKaul (approved):
reminder to add details @ryanbr

PR #2660, opened 2025-10-30
Test disabling ad-shield exception rule

ryanbr:
Disable exception rule and falls back to uAssets rule. See [Slack thread](https://bravesoftware.slack.com/archives/C2FQMN4AD/p1761858211829839?thread_ts=1761830799.905319&cid=C2FQMN4AD) for more context.


ShivanKaul (approved):
Thanks for summary, makes reviews a lot easier and faster.

PR #2666, opened 2025-11-08
localhost-permission-allow-list.txt: pcarrier.com

pcarrier:
Would like this tool to work in Brave. It connects to a local app on 127.0.0.1:12345.


ryanbr:
Should be working without this anymore.

PR #2668, opened 2025-11-13
Add brave-auto-pip scriptlet rules for YouTube and Brave Talk

thypon:
Applies the brave-auto-pip scriptlet to automatically enable picture-in-picture functionality on youtube.com

This provides seamless automatic picture-in-picture experience for video content on these platforms.


ryanbr (approved):
Sure we re-look at this. See if it causes any issues

PR #2683, opened 2025-11-24
Cleanup brave-unbreak.txt

ryanbr:
Scanned with https://www.npmjs.com/package/cleaner-adblock
```
node cleaner-adblock.js --add-www --block-resources --check-dig --input=brave-unbreak.txt
```

* Manually edited/removed rules for this run; 

[dead_domains_2025-11-24T03-03-58.txt](https://github.com/user-attachments/files/23713033/dead_domains_2025-11-24T03-03-58.txt)
[redirect_domains_2025-11-24T03-03-58.txt](https://github.com/user-attachments/files/23713035/redirect_domains_2025-11-24T03-03-58.txt)

### DIFF
--- a/brave-unbreak.txt
+++ b/brave-unbreak.txt
@@ -80,49 +80,42 @@ dzen.ru#@#:matches-path(/articles) div:has(> article > div > [class*="image"]:em
 ! temp fix
 ! https://french-stream.one/films/15122345-anges-amp-cie.html
 ||udeteknvr.com^
-||cuneneifrack.com^
-||peanbowfin.life^
 
 !
 ! .top domains (non-regex)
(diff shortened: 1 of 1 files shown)


ShivanKaul:
Let's revert this PR. This caused a breakage with zoom.us. 

The logic for removing dead domains from the filter list is flawed: the [cleaner-adblock](https://www.npmjs.com/package/cleaner-adblock) tool identifies HTTP-level error and redirecting status codes as evidence for removal, but as we found in the `zoom.us` case, that isn't necessarily true. We had a filter rule for `zoom.us`: `zoom.us##+js(brave-fix)`. This rule was designed to tackle an anti-Brave check done by Zoom on `app.zoom.us` URLs. However, since `zoom.us` now redirects to `zoom.com`, the `cleaner-adblock` tool flagged `zoom.us` domain for removal, even though the sub-domain `app.zoom.us` doesn't redirect and is still used. More problematically, even the other 404ing domains identified in this PR don't necessarily mean that sub-domains of those domains wouldn't still be valid.

The correct way to identify actually-dead domains would be to look for `NXDOMAIN`s when querying for `A` or `AAAA` DNS records. That's the only way we will not have a high false-positive rate for removal of dead domains.

The safest thing to do for now is to revert this cleanup PR and follow-up with a better fix.

PR #2763, opened 2026-02-03
Add _t= for tiktok urls

ryanbr:
https://www.tiktok.com/@in.love.with.angels/video/7569410116240428318?_r=1&_t=ZS-93ccdkqqeqm

`&_t=ZS-93ccdkqqeqm` was requesting GPS when opening on a phone, we can remove this.

### DIFF
--- a/brave-lists/clean-urls.json
+++ b/brave-lists/clean-urls.json
@@ -683,7 +683,8 @@
             "embed_source",
             "refer",
             "referer_url",
-            "referer_video_id"
+            "referer_video_id",
+            "_t"
         ]
     },
     {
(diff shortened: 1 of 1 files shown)


brave-lists/clean-urls.json
```diff
@@ -683,7 +683,8 @@
             "embed_source",
             "refer",
             "referer_url",
-            "referer_video_id"
+            "referer_video_id",
+            "_t"
```
fmarier: Let's move this to the top of the array.

Also, maybe we should add `_r` to the list too? It doesn't seem to be required or do anything.

fmarier (changes requested):

fmarier (approved):

PR #2775, opened 2026-02-11
Add a timestamp and checksum

ryanbr:
./fop --add-checksum=brave-firstparty-regional.txt --add-timestamp=brave-firstparty-regional.txt


ryanbr:
WIP, we'll implement at a later date

PR #2802, opened 2026-03-08
Block Shorts tab on YouTube channel pages

pannya70:
## Description
Adds a filter rule to block the Shorts tab on YouTube channel pages.

## Changes
- Added `youtube.com##yt-tab-shape:has-text(Shorts)` to `brave-lists/yt-shorts.txt` (line 20)

## Motivation
Currently, the Shorts tab on YouTube channel pages (e.g., `youtube.com/@username/shorts`) is not blocked by the existing filter list. Users can still navigate to a channel's Shorts content via this tab.

This is currently the case on both Desktop and Android. I haven't tested it on iOS but I am sure it will be the case there as well. Its pretty easy to replicate:

1. Go to any channel on YouTube using the brave browser such as youtube.com/@arbeanie with the "Block YouTube Shorts" filter applied.
2. You can then see that the YouTube shorts tab is still visisble. Such as:
image

## Implementation
The new rule follows the same pattern as other navigation blocking rules (lines 18-19) by using text-based matching with `:has-text(Shorts)`. This approach is more robust than positional selectors and will work regardless of tab order.

## Testing
- Tested locally using `brave://adblock/` custom filters
- Verified the Shorts tab is hidden on channel pages (e.g., youtube.com/@arbeanie)
- Confirmed other tabs (Videos, Playlists, etc.) remain visible
- No console errors or broken functionality observed

No more Shorts tab:
image
image

### DIFF
--- a/brave-lists/yt-shorts.txt
+++ b/brave-lists/yt-shorts.txt
@@ -17,6 +17,7 @@ youtube.com###items.yt-horizontal-list-renderer > ytd-reel-item-renderer
 ! shorts icons (desktop and mobile)
 youtube.com##.pivot-shorts.pivot-bar-item-tab
 youtube.com##.yt-simple-endpoint[title="Shorts"]
+youtube.com##yt-tab-shape:has-text(Shorts)
 ! Updated rules
 youtube.com##.ytGridShelfViewModelHost
 ! Shorts search header button


ryanbr:
Fixed @pannya70 ?

ryanbr:
Has been fixed, not needed

PR #2806, opened 2026-03-11
Fix Twitch embed issues on `destiny.gg`

vyneer:
One of the pages (https://www.destiny.gg/bigscreen) allows users to embed 3rd party streams to watch; adding the domain to the `checks` list to make the Twitch ones work on Brave.


vyneer:
Seems to be working without this change now, closing.

PR #2812, opened 2026-03-20
Hiding mobile app modal on rotten tomatoes iOS

griffinallen:
Rotten Tomatoes has updated their modal visibility logic here. Now, the modal shows to US users only, on their first visit or 24 hours since the timer was last reset (through a "nextOpen" variable that is set to unix time of now + 24 hours). 

To get around this, we can set the "nextOpen" value to a value indefinitely in the future and set the "wasClosed" value to true. This will prevent the dialog from ever appearing.


ryanbr:
> @ryanbr does it make sense to move these into Mobile Notifications?

Moved for both ios and android, and allows for testing on desktop/mobile UA also @antonok-edm

PR #2821, opened 2026-03-27
Cleanups of tools/debounce_validation/test_debounce_validation.py

ryanbr:
**1st commit:**

- Add unittest.main() so test file can be run directly Without this,  exits silently without running any tests. Tests must be invoked via  instead, which is non-obvious.

**2nd commit:**

- Fix inverted condition in test_regex_path_template_has_redirect_url_template
  (was checking != instead of ==, testing the wrong rules)
- Fix inverted condition and unguarded access in test_redirect_url_template_has_placeholders
  (was checking action != regex-path-template instead of redirect_url_template presence)
- Fix inverted condition in test_redirect_url_template_placeholders_match_capture_groups
  (was skipping regex-path-template rules instead of processing them)
- Use rule.get('action', '') throughout to avoid KeyError if action field is missing

**3rd commit:**

- If a ruleset is missing 'include' or 'exclude', setUp would crash with a KeyError before test_required_fields_present could report the problem cleanly.

### DIFF
--- a/tools/debounce_validation/test_debounce_validation.py
+++ b/tools/debounce_validation/test_debounce_validation.py
@@ -78,15 +78,15 @@ class DebounceTester(unittest.TestCase):
     def test_action_types_are_known(self):
         for i, rule in enumerate(self.raw_debounce):
             self.assertIn(
-                rule['action'], KNOWN_ACTIONS,
-                f"Rule {i} has unknown action '{rule['action']}'"
+                rule.get('action', ''), KNOWN_ACTIONS,
+                f"Rule {i} has unknown action '{rule.get('action', '')}'"
             )
 
     # For regex-path and regex-path-template rules, verify that the param
     # field compiles as a valid regular expression.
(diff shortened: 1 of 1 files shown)


ShivanKaul (changes requested):
The PR description is misleading right now. It claims to fix "inverted conditions" but no conditions are inverted.

The `Use rule.get('action', '') throughout to avoid KeyError if action field is missing` change is a good idea, let's do that.

tools/debounce_validation/test_debounce_validation.py
```diff
@@ -145,3 +145,6 @@ def test_redirect_url_template_placeholders_match_capture_groups(self):
                 f"Rule {i} template references ${max_placeholder} but "
                 f"regex only has {num_groups} capture group(s)"
             )
+
+if __name__ == '__main__':
+    unittest.main()
```
ShivanKaul: missing newline

ryanbr:
Yeah, not needed. Closing

PR #2824, opened 2026-03-30
Add query-filter.json list  

r0hit0303:
Brave currently relies on a hardcoded list of tracking parameters which are baked in the browser: https://github.com/brave/brave-core/blob/386f292095e8dce22b2ce1519761c113eba0aa17/components/query_filter/utils.cc

According to the proposal outlined in the [comment](https://github.com/brave/brave-browser/issues/10188#issuecomment-2415348477), we would slowly move away from that model, to a component updater based approach where the list can be fine-tuned on the fly, making it scalable, robust and adaptive. 

This change adds that initial list. Please see the issue [#10188](https://github.com/brave/brave-browser/issues/10188) for more details.


fmarier (approved):
Looks good.

It's unfortunate that we lose the links to the issues where we analyzed the params, but I don't have a good idea for how to retain these.

r0hit0303:
> Looks good.
> 
> It's unfortunate that we lose the links to the issues where we analyzed the params, but I don't have a good idea for how to retain these.

Maybe we could probably keep track of it as a separate dict inside the json value "params", but that may mean handling things slightly differently on the client side from the rest.  

```json
"params": [
  { "name": "fbclid", "issues": ["https://github.com/brave/brave-browser/issues/4239"] },
  { "name": "gclid", "issues": ["https://github.com/brave/brave-browser/issues/4239"] }
]
```

ryanbr (approved):
Looks fine, good to get away from hardcoded params

PR #2827, opened 2026-04-02
Add rule to remove gs_lcrp from google.com

r0hit0303:
Chromium browsers sends a `gs_lcrp` parameter on searches made to Google via its omnibox. This parameter captures statistical information about omnibox autocomplete impressions which is not relevant for Brave, and instead adds a passive fingerprinting surface.   Please see the [document](https://docs.google.com/document/d/1Ru1qGTSyp3R1bmQYO_aWFoIHEuQW2W1ygBTvKNlFFLs/edit?tab=t.0) for more details on the internal mechanism. 

Fix verified locally on Brave nightly:
https://github.com/user-attachments/assets/6dfe9296-90fb-4e38-856f-6a46ee122498

Resolves https://github.com/brave/brave-browser/issues/49954


r0hit0303:
cc @fmarier @antonok-edm

brave-lists/brave-firstparty.txt
```diff
@@ -1457,6 +1457,7 @@ bloomberg.com##.unsupported-browser-notification-container
 ||warnermediacdn.com/csm*conf_csid=$xhr,removeparam=conf_csid,domain=cnn.com
 ||warnermediacdn.com/csm*prof=$xhr,removeparam=prof,domain=cnn.com
 ||ngtv.io/v2/media/live$xhr,3p,removeparam=ssaiProfile,domain=cnn.com
+||google.com^$removeparam=gs_lcrp
```
fmarier: At the moment, it's under the "remove-params (ads)" heading which isn't right. Let's label it with the actual bug it's fixing:
```suggestion
! omnibox statistics tracking (https://github.com/brave/brave-browser/issues/49954)
||google.com^$removeparam=gs_lcrp
```

ShivanKaul:
Hm but this isn't going to catch all TLDs. For example, if the user's default search engine is google.co.in, this wouldn't work.

ShivanKaul:
We can do `google.*` though! https://github.com/gorhill/uBlock/wiki/Static-filter-syntax#entity

Ugh maybe that won't work for network filters :(

ShivanKaul:
From talking to @antonok-edm `google.*` would still work but it's not an "entity" for network filters the way it is for cosmetic filters. So if a network request to `google.brave.com` had the query param `gs_lcrp` it would get removed.

TBH I think that's fine.

ryanbr (commented):
Comparison from Adguard, they limited to the search; maybe apply the removeparam to the entire google domain is a bit of an overkill. Also covers non dot.com, regional google domains

Sample

```
||google.*/search$removeparam=gs_lcrp
```

If we go without the wildcard, maybe use www.google.com/search  to limit to a www record. maybe cover more than .com? So multiple lines for each regional google search

```
||www.google.com/search$removeparam=gs_lcrp
||www.google.com.au/search$removeparam=gs_lcrp
||www.google.co.nz/search$removeparam=gs_lcrp
||www.google.ca/search$removeparam=gs_lcrp
||www.google.co.uk/search$removeparam=gs_lcrp
||www.google.de/search$removeparam=gs_lcrp
||www.google.fr/search$removeparam=gs_lcrp
||www.google.es/search$removeparam=gs_lcrp
||www.google.pt/search$removeparam=gs_lcrp
||www.google.ie/search$removeparam=gs_lcrp
||www.google.dk/search$removeparam=gs_lcrp
||www.google.nl/search$removeparam=gs_lcrp
||www.google.pl/search$removeparam=gs_lcrp
```

ryanbr (approved):

PR #2852, opened 2026-04-20
Fix IOS UA checks on novey.com.pa

ryanbr:
Reported ios issues on https://www.novey.com.pa/productos-electricos-y-ferreteria/cajillas-tapas-tomas-y-switches/conectores-enchufes-y-terminales


StephenHeaps:
Just trying to confirm this is related to iOS UA, as I get 403 in Safari, Chrome and Firefox as well.

StephenHeaps (approved):

PR #2853, opened 2026-04-21
Fixes UA checks on jewelosco.com

ryanbr:
Image from iOS

### DIFF
--- a/brave-lists/brave-checks.txt
+++ b/brave-lists/brave-checks.txt
@@ -289,6 +289,7 @@ instantconsult.com.au
 zoro.to
 healthrangerstore.com
 saramart.com
+jewelosco.com
 html-code-generator.com
 support-lock.com
 fordeal.com


brave-lists/brave-checks.txt
```diff
@@ -289,6 +289,7 @@ instantconsult.com.au
 zoro.to
 healthrangerstore.com
 saramart.com
+jewelosco.com
```
ShivanKaul, line 292:
```suggestion
! iOS UA check
jewelosco.com
```

ShivanKaul: we should mark these in the file so we can easily clean up later. I think a "comment" like this would just be a no-op in brave-core because it wouldn't pass the `GetDomainAndRegistry` but wondering if it would fail a test on crx-packager.

ShivanKaul: Looks like not. So we should be good to add a comment like this. For bonus points we could strip out all the comments before shipping to clients.

StephenHeaps (approved):

PR #2880, opened 2026-05-11
Implementing brave-youtube-playback-speed-fix to fix youtube playback speed issues

griffinallen:
This fixes a bug where only changing playback speed with the slider persisted between videos. When adjusting the playback speed with the buttons they provide, the speed does not persist for the next video. This is confusing for users and is fixed in this commit.

The scriptlet related to this fix is found in this PR: https://github.com/brave/adblock-resources/pull/320


ShivanKaul (approved):
Same www prefix comment here. Would be good to get @ryanbr  eyes on this when he's on to see if we can fold into upstream

PR #2891, opened 2026-05-22
Replace sec-team with a more targeted team for debounce list

fmarier:
Also add the new query-filter list to the same category


kdenhartog (approved):
makes sense to me so we can build around the specific expertise for this

PR #2900, opened 2026-05-28
Add validation script for query string filter

fmarier:
This ensures that the conditional entries are not added to the main list.

See https://github.com/brave/brave-core/pull/36810/files#r3319995240.

### DIFF
--- a/tools/validate_query_filter.py
+++ b/tools/validate_query_filter.py
@@ -1,0 +1,104 @@
+#!/usr/bin/env python3
+
+"""Validate that no parameters in brave-lists/query-filter.json overlap with
+the keys of kConditionalQueryStringTrackers in brave-core's utils.cc.
+
+Those parameters are stripped conditionally by brave-core (only when the URL
+spec does NOT match a regex). Listing them in query-filter.json — where they
+would always be stripped — defeats the conditional behavior.
+"""
+
+import json
+import os
+import re
+import sys
+import urllib.request
+
+
+UTILS_URL = os.environ.get(
+    "UTILS_URL",
+    "https://raw.githubusercontent.com/brave/brave-core/refs/heads/master"
+    "/components/query_filter/browser/utils.cc",
+)
+QUERY_FILTER = "brave-lists/query-filter.json"
+
+
+def fetch_utils_cc(url):
+    with urllib.request.urlopen(url) as resp:
+        return resp.read().decode()
+
+
+def extract_conditional_keys(source):
+    """Return the sorted set of keys defined in kConditionalQueryStringTrackers.
+
+    Entries in the source look like:  {"key", "value"},
+    """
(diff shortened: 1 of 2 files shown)


tools/validate_query_filter.py
```diff
@@ -0,0 +1,104 @@
...
+UTILS_URL = os.environ.get(
+    "UTILS_URL",
+    "https://raw.githubusercontent.com/brave/brave-core/refs/heads/master"
+    "/components/query_filter/browser/utils.cc",
+)
+QUERY_FILTER = "brave-lists/query-filter.json"
+
+
+def fetch_utils_cc(url):
+    with urllib.request.urlopen(url) as resp:
```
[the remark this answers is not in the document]
fmarier: It's a constant.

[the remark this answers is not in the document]
fmarier: It's a constant.

tools/validate_query_filter.py
```diff
@@ -0,0 +1,104 @@
...
+    for rule in data:
+        params.update(rule.get("params", []))
+    return params
+
+
+def main():
+    repo_root = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+    query_filter_path = os.path.join(repo_root, QUERY_FILTER)
+
+    source = fetch_utils_cc(UTILS_URL)
```
r0hit0303: Should we add a check here to make sure `/components/query_filter/browser/utils.cc` has not been moved/renamed? similar to how you are handling for `conditional_keys` parsing. Right now, it would be implicitly captured by ` if not conditional_keys:` in L67, but I'm afraid that the current message wouldn't be super clear if the file had been moved/renamed. wdyt?

tools/validate_query_filter.py
```diff
@@ -0,0 +1,104 @@
...
+    # good enough for the compact format brave-core uses.
+    match = re.search(r"\bkConditionalQueryStringTrackers\s*=", source)
+    if match is None:
+        return []
+    start = match.end()
+    end = source.find("});", start)
+    if end == -1:
+        return []
+    block = source[start:end]
+    return sorted(set(re.findall(r'\{\s*"([^"]+)"\s*,', block)))
```
r0hit0303, line 51: Should we add `block = re.sub(r'//[^\n]*', '', block)` to remove comments so we don't accidentally match with something inside it?

r0hit0303 (approved):
LGTM. Everything looks great!

tools/validate_query_filter.py
fmarier: Good idea. Done in b5bf9e2524bf5e56a27b3a8552edced96cb7edd0

tools/validate_query_filter.py
fmarier: Added in 1d76295cd36ad8d72acbc85a852c3a53fd7bb8e7.

PR #2901, opened 2026-05-29
Fixing Apple Pay popup on camh.ca

griffinallen:
The apple pay popup was not visible on this site. To fix this, it was found that both app.blackbaud.net and stripe.com needed to be allowlisted in order for the popup to display properly.

I attempted this with uBO on Safari and couldn't reproduce. This issue seems to be isolated to Brave on iOS


griffinallen:
This was fixed by @ryanbr in the following Easylist change:
https://github.com/easylist/easylist/commit/8f70daa5

ryanbr:
We'll close this

PR #2905, opened 2026-06-01
Re-adding SABR YT fix to experimental list

griffinallen:
This PR reintroduces the SABR YT fix to the experimental list. This fix was removed due to speculation that it was causing a more widespread Youtube issue. We have determined the true cause of this issue and it is unrelated to this. 

I would like to continue monitoring user feedback with this in experimental with hopes that we can eventually bring this into the default filter list.


ryanbr:
Is it still an issue currently? is it needed?

griffinallen:
We still have a small delay (3-5 seconds) upon loading youtube videos. With the SABR fix, this delay shrinks significantly and is less than 1 second most of the time.

PR #2915, opened 2026-06-09
Fixing anti-adblock on winfuture.de

griffinallen:
This site is using adshield with html-load to serve ads and detect if any of them are blocked. I have found that allowing the html-load scripts, xhr requests, fetch requests and any document requests from html-load/feed will make the page run as expected.

@ryanbr - if there is a more appropriate solution here, please suggest it.

Fixes https://github.com/brave-experiments/spiking-webcompat-reports/issues/2023


ryanbr:
Would need some time to investigate this, but lets have it working for the time being

PR #2942, opened 2026-06-30
Fixing anti-adblock on flacdownloader.com

griffinallen:
This site has an anti-adblock message that appears under the condition that navigator.brave exists. To bypass this, we need to implement our brave-fix on this site

Related to https://github.com/brave-experiments/spiking-webcompat-reports/issues/2082


ryanbr:
Has other changes included here, maybe rebase? @griffinallen

griffinallen:
@ryanbr - rebased

griffinallen:
This fix is no longer necessary

PR #2959, opened 2026-07-13
[clean-urls] Sort the global list in alphabetical order

r0hit0303:
This is a purely **mechanical** change which sorts the existing blocked list of parameters from the global filter. This helps to keep the list in order to facilitate code readability.

The two lists are verified to have the same contents here: https://godbolt.org/z/6c7ojEbcz


r0hit0303:
Closing this PR since @ryanbr advised offline that sorting the parameters could affect the git history.

PR #2968, opened 2026-07-25
Add ksp.co.il to brave-checks.txt

venusddev:
ksp.co.il now blocks brave browser


venusddev:
needs to be added on brave-fix now too, since they are trying alternative methods to detect

PR #3009, opened 2026-09-16
Restore some sites for iOS UA block

StephenHeaps:
`edsby.com` (unsupported notice with click-through) and `wenku8.net` (blocked access) are having issues on Brave iOS with the updated iOS user agent. Fine with Safari Mobile, so adding to `brave-checks.txt` to store block bypass.


griffinallen (approved):

ShivanKaul (approved):
Ah, looks like these ones fail even on the [new iOS UA](https://docs.google.com/spreadsheets/d/1ISgpLZjO4bvHll9iNH3qOQqqt4kXBkZzozbgvqOs26c/edit?gid=0#gid=0). Let's ship this!

PR #3010, opened 2026-09-17
time-limited investigation of cookie banners

iambrianfung:
see: https://github.com/brave/reviews/issues/2372


ryanbr:
How would user want to disable the rules , if they wanted too, $important:
 
`||sdk.privacy-center.org^$script,important,domain=oscaro.com|boulanger.com`

Not sure about to get around:
```
oscaro.com,boulanger.com#@##didomi-host
oscaro.com#@#+js(trusted-click-element, .didomi-continue-without-agreeing, , 1000)
```

Edit: Could be change too: By changing the delay
```
oscaro.com##+js(trusted-click-element, .didomi-continue-without-agreeing, , 1500)
```

If its just a test, its fine. But if going forward the user should be able to disable these.

ryanbr (approved):

